PR #21, opened 2025-10-13
Fix/reset wifi credentials

tadelv:
HDS homepage was missing a "Reset WiFi settings" button, leaving users unable to reset WiFi back to stock.

This PR adds a "Reset WiFi" button to the HDS homepage which calls the /setup/wifi endpoint with an empty ssid parameter, effectively preventing HDS to try and connect to an erroneous WiFi on setup.

I've also added the call to `esp_reset()` in the function that handles WiFi parameters application, streamlining the WiFi setup as users no longer need to reset the scale manually.

### DIFF
--- a/include/menu.h
+++ b/include/menu.h
@@ -3,6 +3,7 @@
 
 #include "esp32-hal.h"
 #include "parameter.h"
+#include "wifi_setup.h"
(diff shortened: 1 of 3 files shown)


tadelv:
I want to add an HDS Setup menu entry for resetting the WiFi as well, so that all the options are covered

PR #27, opened 2025-11-21
Fix/wifi and hds setup

tadelv:
- Perform Wifi/Server setup as a separate task to avoid hangup on start when WiFi is not present or misconfigured.

- Enable Wifi only if BLE is also enabled

- Blink wifi icon until HDS is connected or Access Point is established

- Fix crash that would sometimes occur during calibration

### DIFF
--- a/git_rev_macro.py
+++ b/git_rev_macro.py
@@ -1,0 +1,9 @@
+import subprocess
+
+revision = (
+    subprocess.check_output(["git", "rev-parse", "HEAD"])
+    .strip()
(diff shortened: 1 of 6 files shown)


tadelv:
Added:

- Blinking of wifi icon (chain link) when in process of searching/connecting to saved Wifi network
- On About screen, display build date and commit hash, added a py script that gets the commit and is run by pio on build time.
- Use a fixed length char buffer for dtostrf in load cell calibration (should fix crash sometimes happening when calibrating)

Issue #22, opened 2025-10-17
Timer is started when exiting HDS Setup mode


tadelv:
When leaving HDS Setup mode and returning to home screen, the timer is started. Probably due to firmware registering the button tap for both exiting the menu and starting timer at the same time.

Sofronio:
I can confirm this is a long existing known issue. I'll try fix it.

Sofronio:
Fixed here: [https://github.com/decentespresso/openscale/tree/Sofronio_exitMenuTimerTriggered_fix](https://github.com/decentespresso/openscale/tree/Sofronio_exitMenuTimerTriggered_fix)

tadelv:
Thank you @Sofronio !

PR #28, opened 2025-12-04
fix: prevent scaleTimer trigger immediately after menu exit

tadelv:
- Add t_menuExitTime variable to track menu exit time
- Implement 500ms protection period after exiting menu
- Prevent scaleTimer() from triggering during menu transition


tadelv:
closes #22

PR #35, opened 2026-01-30
removed no weight detected

tadelv:
No weight detected used to prevent user from doing a wrong calibration. But when user change another load cell, or user accidentally performed a successful wrong calibration, the cal_value will be very wrong, thus enter the "if (abs(d_weight) < 5)" and oled will show no weight detected, and block user from recalibration.

### DIFF
--- a/include/menu.h
+++ b/include/menu.h
@@ -745,25 +745,6 @@ void calibration(int input) {
         }
         Serial.print("weight is ");
         Serial.println(d_weight);
-//         if (abs(d_weight) < 5) {
-//           u8g2.firstPage();
-//           u8g2.setFont(FONT_S);
-//           do {
-//             // 2行
-//             // FONT_M = u8g2_font_fub14_tn;
-//             u8g2.drawUTF8(AC((char *)"No weight detected"), AM(),
-//                           (char *)"No weight detected");
-//           } while (u8g2.nextPage());
-// #ifdef BUZZER
-//           buzzer.off();
-// #endif
-//           delay(1000);
-//           // reject the weight and exit
-//           i_button_cal_status = 0;
-//           b_calibration = false;
-//           b_menu = true;
-//           return;
-//         }
         scale.refreshDataSet();  // refresh the dataset to be sure that the known
                                  // mass is measured correct
 
--- a/include/menu.h
+++ b/include/menu.h
@@ -745,25 +745,25 @@ void calibration(int input) {
         }
         Serial.print("weight is ");
         Serial.println(d_weight);
-        if (abs(d_weight) < 5) {
-          u8g2.firstPage();
-          u8g2.setFont(FONT_S);
-          do {
-            // 2行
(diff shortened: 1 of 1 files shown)


tadelv:
@Sofronio you can also delete code instead of commenting it out.
We have git to keep the history

decentjohn:
Customer reports the problem not entirely solved.

> My half decent scale arrived yesterday and was working fine for about 6 hours and then on the next turn on - it just reads 0 constantly. Trying to recalibrate constantly shows a message that says the device did not detect a weight. Any ways to resolve the issue?

> john here -- I do know that the checking for a weight during calibration, was removed in one of the more recent hds firmware updates. Ah yes I see that noted here in 3.0.5 [github.com/decentespres…](https://github.com/decentespresso/openscale/releases/tag/v3.0.5)
> User Avatar
> 
> whats happened is oddly your calibration was lost and then set to something wrong, and that's preventing recalibration. This new fw should fix that. Sorry about this!

> Hi John - thank you for your reply. I changed the firmware to 3.0.5 yesterday evening and I do get a message saying that it was calibrated with the 50gram weight 
> 
> But even after doing that step - the scale constantly weighs at 0 grams
> 
> I tried 3.0.5,3.0.4, 3.0.3, 3.0.2, 3.0.1
> but everything constantly shows 0 grams. Wi-Fi connection and the button responsive all seem fine
> Cliff 
> 
> Could it be anything else?
> Cliff 

> I think the patch is  incomplete 
> 
> You have a block of code that’s been commented out that is the var known_mass to be 0 in all cases and then any attempts at calibration will divide by 0 - which might be attributing to why my scale constantly reads 0
> 
> 0850: //         bool valid_mass = false;
> 0851: //         for (int i = 1; i <= 40; i++) {
> 0852: //           // check from 50g to 2kg
> 0853: //           if (current_weight > i * 50 - 2 && current_weight < i * 50 + 2) {
> 0854: //             // check if the weight is within 2g
> 0855: //             known_mass = i * 50;
> 0856: //             valid_mass = true;
> 0857: //             break;
> 0858: //           }
> 0859: //         }
> 0861: //         if (!valid_mass) {
> 0862: //           char buffer[50];
> 0863: //           snprintf(buffer, sizeof(buffer), "%f.0g weight detected",
> 0864: //                    current_weight);
> 0865: //           ...
> 0879: //           b_calibration = false;
> 0880: //           return;  // exit calibration
> 0881: //         }

cliffpham:
Found a bug in 3.0.5: in menu.h calibration state 4 (smart/any-weight), known_mass is left at 0 because weight-detect code is commented out. That makes getNewCalibration() divide by 0 → calFactor becomes INF and calFactorRecip becomes 0, so weight reads 0 forever. Manual 50g recalibration won’t recover because it multiplies by existing INF calFactor. Need EEPROM cal value reset + boot-time isfinite/range check

PR #50, opened 2026-05-08
Fix drift compensation reset state

skialpine:
## Summary

Fixes the drift compensation state reset in `pureScale()`.

The invalid-diff branch previously declared a second block-local `static int f_similar_diff_count`, so it reset a different counter than the one used by the compensation path. This moves the counter and last-diff state to the `pureScale()` scope and resets the active state when the diff is outside the drift-compensation window.

## Validation

- `platformio run`
- Flashed/tested the same source change locally before opening this standalone PR

## Notes

This PR is intentionally scoped to the drift compensation reset bug and is based directly on `develop`.


tadelv:
Thanks @skialpine !

PR #52, opened 2026-05-15
Fix intermittent BLE connect/notify drop + graceful radio shutdown

tadelv:
## Summary
- Fix intermittent bug where BLE icon kept blinking and weight notifications stopped, even though tare/shutdown still worked — `MyServerCallbacks` now uses NimBLE's connection-aware overloads and ignores stale disconnects from old `conn_handle`s
- Centralize graceful BLE + WiFi + web server teardown into `esp32_sleep()` and `reset()` so every shutdown path drops the radios cleanly

## Why
The bug only appears under NimBLE — a silent regression from `e748fda`, which bumped arduino-esp32 3.2 → 3.3 and thereby switched the esp32s3 BLE stack from Bluedroid to NimBLE. The pre-existing 1-arg connId-blind callbacks were fine under Bluedroid, but under NimBLE a transient central drop+reconnect can land a stale disconnect for an old handle *after* the new connection's `onConnect`, clearing `deviceConnected`/`connId` while the new link is alive. Ungraceful shutdown (no `pServer->disconnect()` anywhere, peer learns via supervision timeout) is a likely trigger for that drop+reconnect dance.

Runtime-verified with Decent.app / ReaPrime (connect / power-off cycles plus an overnight soak).


tadelv:
Huge commit due to change of how line endings are stored (CRLF->LF). The real fix is in last three commits.

PR #54, opened 2026-05-21
feat: WiFi scale support — WebSocket control + events

skialpine:
## Summary

Brings the on-device `/snapshot` WebSocket up to feature parity with BLE so the web UI and a separate app (phone/desktop) can both drive the scale.

**New protocol** ([c0eb374](https://github.com/skialpine/openscale/commit/c0eb374))
- One-way `{"grams","ms"}` weight stream extended into a full **control + event** protocol.
- Event broadcasts: button presses (`sendWebsocketButton`) and power-off reasons (`sendWebsocketPowerOff`), wired into the same call sites as the BLE/USB equivalents.
- Status frames (`sendWebsocketStatus` / `sendWebsocketStatusAll`) publishing battery %, voltage, charging, timer, display, soft-sleep, LED, and rate state — broadcast periodically every 5 s after `events on`, and on demand.
- Commands accepted as text or JSON: `status`, `events on/off`, `tare`, `timer start/stop/zero`, `led <r,g,b>|off`, `display on/off`, `low_power on/off`, `sleep on/off`, `power off`.
- Selectable weight rate: 2 Hz / 5 Hz / 10 Hz via `rate <hz>`, `interval <ms>`, or JSON (`rate_hz` / `hz` / `rate` / `interval_ms` / `{command:rate,value:…}`).

**Freeze fix** ([1051681](https://github.com/skialpine/openscale/commit/1051681))
- The WS event handler runs on the AsyncTCP task, so calling `u8g2.setPowerSave/setContrast` or `digitalWrite(PWR_CTRL,…)` directly raced the main loop's OLED draws and stalled the device for ~4 s under feature-test load.
- Added a `portMUX`-protected `wsPendingMask` drained early in `loop()`. Hardware ops deferred; state flags updated synchronously so status frames stay accurate.
- Tightened `include/webserver.h`: handlers registered once across stop/start (`server.end()` doesn't clear the handler list), `server.begin()` moved to the end, removed a duplicate `addHandler(&websocket)`.
- WS DATA callback: `String((const char*)data, len)` instead of O(N²) char-by-char concat; only accepts complete unfragmented text frames.
- `static unsigned long lastUpdate` (was signed → wrong across `millis()` rollover).
- Documented in `wifi_setup.cpp` that `WiFi.setSleep(false)` is the **wrong** call with BLE active (measured: 8 % packet loss, HTTP all-timeout because BT+WiFi share the 2.4 GHz radio and the coex layer needs WiFi's sleep windows for BT slots).

**Review feedback** ([74d7f0f](https://github.com/skialpine/openscale/commit/74d7f0f))
- Mask collapse fix: `wsPendingMask |= bit` let opposing actions accumulate, so `display off` then `display on` could end up off while status reported on. Added `wsReplacePending(set, clear)` for the three mutually-exclusive pairs.
- Marked cross-task globals (`weightWebsocketNotifyInterval`, `b_websocketEventsEnabled`, `b_softSleep`, `b_u8g2Sleep`, …) `volatile` for dual-core ESP32-S3 cache coherence.
- Cast `stopWatch.elapsed()` to `(unsigned long)` for `%lu`. Fixed two stale doc comments (`wsQueuePending` and `processWsPendingCmds`). README documents the JSON rate forms (`rate_hz`, `hz`, `interval_ms`) that were accepted but undocumented.

**Multi-client** ([7a3b51c](https://github.com/skialpine/openscale/commit/7a3b51c))
- Dropped the 1-client middleware so the web UI and an app can both connect to the same scale.
- `cleanupClients()` now uses the library default (`DEFAULT_MAX_WS_CLIENTS` = 8 on ESP32).
- `WS_EVT_DISCONNECT` only resets the shared rate/events state when `server->count() == 0` (last client out) — otherwise one client leaving would wipe the other's negotiated state.

## Verified on device
- 60/60 WebSocket feature regression PASS (`status` / `events` / `tare` / `timer` / `led` / `display` / `low_power` / `sleep` / rate forms / unknown-input handling / soft-sleep round trip).
- 120 s freeze stress (10 Hz WS + HTTP every 1 s + ~5 cmd/s feature chaos): max weight gap **0.60 s** (was **4.18 s** pre-fix), 0 HTTP failures, 0 slow.
- Multi-client 5/5 PASS: two simultaneous handshakes both succeed; both clients stream weight at the negotiated rate; both see the periodic status broadcast; disconnecting one client leaves the other's stream and state intact.


skialpine:
### Code review

Found 1 issue, fixed in [74d7f0f](https://github.com/skialpine/openscale/commit/74d7f0f):

1. **Deferred-action mask collapses opposing commands.** `wsPendingMask` was set with `|=` only, so if a client sent `display off` then `display on` before the main loop drained, both bits ended up set and `processWsPendingCmds` ran them in source order — final hardware was off while `b_u8g2Sleep` reported on, leaving the status frame inconsistent with the device. Same shape for `low_power on/off` and `sleep on/off`. Fixed by adding `wsReplacePending(set, clear)` and atomically superseding the opposing bit at the three pair sites.

   https://github.com/skialpine/openscale/blob/7c533a241af23df7654c7e3cc2f087704ce1580b/src/hds.ino#L496-L508

   https://github.com/skialpine/openscale/blob/7c533a241af23df7654c7e3cc2f087704ce1580b/src/hds.ino#L751-L803

---

<details>
<summary>5 below-threshold issues (50-74), all addressed in 74d7f0f</summary>

- [Worth a look] **Missing `volatile` on cross-task globals** — `weightWebsocketNotifyInterval`, `b_websocketEventsEnabled`, `b_softSleep`, `b_u8g2Sleep` etc. are written from the AsyncTCP task and read from the main loop on the other core. Without `volatile`, the compiler may cache stale values in a register across the WS gate check. Fixed by marking them `volatile`. (score: 70)

  https://github.com/skialpine/openscale/blob/7c533a241af23df7654c7e3cc2f087704ce1580b/include/parameter.h#L20-L43

- [Worth a look] **README missing JSON rate forms** — code accepts `{ "rate_hz": N }`, `{ "hz": N }`, `{ "interval_ms": N }` but only the string-key forms were documented. Added to the examples. (score: 65)

  https://github.com/skialpine/openscale/blob/7c533a241af23df7654c7e3cc2f087704ce1580b/README.md#L66-L75

- [Worth a look] **`wsQueuePending` doc-comment listed `stopWatch` as something that must not be touched from the AsyncTCP task, but the `timer` branch does deliberately touch it** (it's just bool/uint32 writes — safe). Removed stopWatch from the comment. (score: 60)

  https://github.com/skialpine/openscale/blob/7c533a241af23df7654c7e3cc2f087704ce1580b/src/hds.ino#L486-L493

- [OK to skip] **`processWsPendingCmds` `Hardware ops only` comment was contradicted by the `b_powerOff = true` write inside it.** Comment updated to explain why `b_powerOff` is set here on purpose (sequence after the centralized shutdown chain). (score: 55)

  https://github.com/skialpine/openscale/blob/7c533a241af23df7654c7e3cc2f087704ce1580b/src/hds.ino#L513-L518

- [OK to skip] **`%lu` ↔ `uint32_t` type mismatch** on `stopWatch.elapsed()` in both status-frame printfs. Works on ESP32 (both 32-bit) but technically UB by the C standard. Added explicit `(unsigned long)` cast. (score: 50)

  https://github.com/skialpine/openscale/blob/7c533a241af23df7654c7e3cc2f087704ce1580b/src/hds.ino#L584-L589

</details>

Verification after these fixes: 60/60 WS feature regression PASS, 120s freeze stress max weight gap 0.60s (4.18s pre-fix), 0 HTTP failures.

False positives discarded: an `events on` → invalid claim (`command == "events"` is exact equality, not a prefix match, so the first call falls through and the split-and-retry path handles it correctly); a `sendWebsocketPowerOff` ordering risk (it fires before `esp32_sleep()` which is where `stopWebServer()` runs, so the event is delivered before WS teardown).

Generated with [Claude Code](https://claude.ai/code)

src/wifi_setup.cpp
```diff
@@ -51,6 +52,10 @@ void connectToWifi() {
 
   WiFi.begin(params.getSSID(), params.getPass());
   WiFi.setTxPower(WIFI_POWER_18_5dBm);
+  // Leaving WiFi modem-sleep at the Arduino default (WIFI_PS_MIN_MODEM):
```
tadelv: this comment here seems redundant?

src/hds.ino
```diff
@@ -367,6 +373,578 @@ void button_init() {
...
+    command.trim();
+    command.toLowerCase();
+    String action = "";
+    if (doc["action"].is<const char *>()) {
+      action = doc["action"].as<String>();
+    }
+    if (command == "rate" && doc["value"].is<const char *>()) {
+      return setWebsocketRateFromValue(client, doc["value"].as<String>());
+    }
+    if (command == "led" &&
```
tadelv: LedRGB seems to have snuck in somehow, since it doesn't really do anything, could you remove all the references and code related to it?

tadelv (commented):
I'd like ask you to move the websocket code into a file that is not `hds.ino`. Some separation of concerns would be nice to have.

src/hds.ino
```diff
@@ -1404,20 +1993,31 @@ void loop() {
...
           ElegantOTA.loop();
-          static long lastUpdate = 0;
+          static unsigned long lastUpdate = 0;
           unsigned long current = millis();
-          if (current - lastUpdate > 500) {
+          if (current - lastUpdate >= weightWebsocketNotifyInterval) {
             if (websocket.availableForWriteAll() > 0) {
               if (!b_adc_recovery_active) {
-                websocket.printfAll("{ \"grams\": %.2f, \"ms\": %lu }", f_displayedValue, current);
+                websocket.printfAll("{\"grams\":%.2f,\"ms\":%lu}", f_displayedValue, current);
```
tadelv: `availableForWriteAll()` returns the **minimum** writable across all clients, so this all-or-nothing gate couples every client to the slowest one: if any single client's TX buffer is full (slow/dead socket), the weight broadcast — and the periodic status below — is skipped for *everyone*. That defeats the PR's multi-client independence goal. Consider iterating clients and sending per-client (check each `client->canSend()`) instead of gating the shared `printfAll`.

src/hds.ino
```diff
@@ -380,23 +958,29 @@ void _wifi_init(void *args) {
...
-      Serial.println(msg);
-      if (msg == "tare") {
-        requestRemoteTare();
+      // Only handle complete, unfragmented text frames. Fragmented or binary
+      // frames would corrupt the parsers below.
+      if (info->final && info->index == 0 && info->len == len && info->opcode == WS_TEXT) {
+        String msg((const char *)data, len);
+        Serial.print("Websocket recv: ");
+        Serial.println(msg);
+        handleWebsocketRateCommand(client, msg);
```
tadelv: Trust-model note: pre-PR this socket was read-only weight + `tare`. It now accepts `power off`, `soft_sleep on`, `display off`, `led`, `timer`, etc. with no authentication. Anyone on the LAN can power the scale off (and `power off` can't be woken remotely) or sleep it — easy griefing/DoS. If a trusted-LAN assumption is intended, please state it explicitly; otherwise consider gating the destructive commands.

src/hds.ino
```diff
@@ -367,6 +373,578 @@ void button_init() {
...
+  }
+
+  if (command == "tare") {
+    requestRemoteTare();
+    sendWebsocketStatus(client, "ok");
+    return true;
+  }
+
+  if (command == "timer") {
+    // StopWatch is just bool/uint32 state — safe to mutate from the AsyncTCP
```
tadelv: "safe to mutate from the AsyncTCP task" isn't quite right. `StopWatch` is multi-field (running flag + start timestamp + accumulator), and it's also mutated from the main loop (`hds.ino` shot logic), BLE (`ble.h`), and USB (`usbcomm.h`). Concurrent `start/stop/reset` here plus `elapsed()/isRunning()` reads in the status frame can tear or read mid-update across cores. `requestRemoteTare()` already uses a portMUX for exactly this reason — the timer path warrants the same guarding rather than an assumption that bool/uint32 writes are atomic-as-a-group.

src/hds.ino
```diff
@@ -367,6 +373,578 @@ void button_init() {
...
+      b_u8g2Sleep = true;
+      wsReplacePending(WSP_DISPLAY_OFF, WSP_DISPLAY_ON);
+      sendWebsocketStatus(client, "ok");
+      return true;
+    }
+    sendWebsocketStatus(client, "invalid");
+    return true;
+  }
+
+  if (command == "low_power") {
```
tadelv: `low_power on` defers `setContrast(0)`, but the button-press path sets `setContrast(255)` (`hds.ino:94`). After any button press the contrast reverts to max while `b_websocketLowPowerEnabled` stays `true`, so the status frame reports `low_power:true` that no longer matches the hardware.

src/hds.ino
```diff
@@ -367,6 +373,578 @@ void button_init() {
...
+    sendWebsocketStatus(client, "invalid");
+    return true;
+  }
+
+  if (command == "sleep" || command == "soft_sleep") {
+    if (action == "on") {
+      Serial.println("Websocket soft sleep on detected.");
+      // Set state flags synchronously so status reflects the requested mode.
+      // u8g2 + GPIO power-rail writes are deferred to the main loop.
+      b_softSleep = true;
```
tadelv: `soft_sleep on` calls `setPowerSave(1)` (display off) but never sets `b_u8g2Sleep = true`. Since the status frame derives `display_on` from `b_u8g2Sleep`, it reports `display_on:true` while the OLED is physically off during soft sleep. `soft_sleep off` *does* set `b_u8g2Sleep = false`, so the pair is asymmetric — set `b_u8g2Sleep = true` here too.

tadelv (commented):
Strict behavioral pass — four findings beyond the cleanup/structure notes I left inline. None block the protocol design, but #1 and #2 undercut the stated multi-client goal/threat model, and #3–#4 make status frames report state that isn't real.

src/hds.ino
tadelv: Fine to keep `printfAll`. Agreed the per-client `getClients()` iteration races the library's client list, and the slow-client coupling I flagged is an edge case — on a home LAN we rarely expect more than one client, and dropping the `availableForWriteAll` gate removes the all-or-nothing behavior anyway. No change needed here.

Worth revisiting in the future if the library exposes a lock-safe per-client send (or the client-list race goes away), at which point per-client backpressure would be a clean win for the multi-client case.

tadelv (approved):
All review points addressed — thanks for the quick turnaround:
- WS code moved out of `hds.ino` into `include/websocket.h`
- LedRGB removed
- `soft_sleep on` now sets `b_u8g2Sleep` (status frame truthful)
- button path clears `b_websocketLowPowerEnabled` so `low_power` status stays accurate
- StopWatch timer ops deferred under `wsPendingMux` (no cross-core tear)
- trust model documented explicitly (unauthenticated, trusted-LAN only)

Broadcast stays on `printfAll` — agreed, per the thread above. Noted there that per-client backpressure is worth revisiting if the library ever exposes a lock-safe per-client send.

LGTM.

PR #55, opened 2026-05-21
docs: add CLAUDE.md with project gotchas

skialpine:
## Summary
Add a `CLAUDE.md` at the repo root documenting project-specific knowledge that isn't obvious from the source: build/flash commands, code layout (the headers-with-implementations pattern, where `setup()` and `loop()` live, what `config.h` does), the AsyncTCP-task vs main-loop threading model and the `wsPendingMask` deferral pattern, why `WiFi.setSleep(false)` is the wrong call with BLE active, ESPAsyncWebServer init ordering, the macOS `curl -4` AAAA workaround, AP-mode fallback credentials, a "when X is broken, check Y" table, naming conventions, and a `.gitattributes` LF-normalization note for bisects.

Distilled from the debugging session that produced #54. Split out as its own PR so #54 stays focused on the firmware change.

## Merge order

**Land #54 first.** This PR documents the WebSocket protocol, deferral pattern (`wsReplacePending` / `processWsPendingCmds` / `WSP_*` bits), WS frame-guard snippet, and corresponding line refs that are introduced by #54. Against bare `main` the references look forward; against `main` after #54 lands they're accurate.

## Test plan
- [ ] Read through once after #54 has merged — no commands to run; pure docs

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,142 @@
+# CLAUDE.md — Half Decent Scale firmware
+
+Espresso-scale firmware for the HDS hardware: ESP32-S3 + load-cell amplifier (ADS1232 or HX711) + 128x64 OLED + BLE + WiFi. Built with PlatformIO + Arduino framework. The on-device web app at `/` talks to the firmware over a single `/snapshot` WebSocket.
+
+## Quick reference
+
+```sh
+# All commands run from the repo root.
+pio run -e esp32s3                                                # build
+pio run -e esp32s3 -t upload --upload-port /dev/cu.wchusbserial110 # flash firmware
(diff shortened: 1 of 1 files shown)


tadelv (approved):
Solid knowledge transfer doc. Captures the real footguns from #54 — threading model, WiFi/BLE coexistence, ESPAsyncWebServer gotchas. The safety matrix and troubleshooting table are excellent. One minor note: the hardcoded `/opt/homebrew/Cellar/platformio/6.1.19_1/` path is machine-specific and will break on version bumps — consider `brew --prefix platformio` or moving pyserial to a troubleshooting footnote. Not blocking.

PR #56, opened 2026-05-23
WiFi reliability + true 10 Hz output across WiFi/BLE/USB

skialpine:
## Summary
Hardens the WiFi/network path and unifies weight output so the scale reliably delivers 10 Hz over WiFi (plus USB/BLE) under real-world load. Five focused commits:

- **WiFi resilience** — reconnect supervisor with exponential backoff (recovers from STA deauth / `WL_STOPPED` that previously left WiFi dead until reboot), disconnect-reason + heap logging, mDNS re-advertise on reconnect, and a heap watchdog (reboot on sustained OOM that wedges the lwIP stack).
- **Network/HTTP** — raise the AsyncTCP ACK timeout on the WS connection to ride out transient BT/WiFi coexistence radio stalls (the ~1–2 min "clean disconnect" drops); HTTP admission control that sheds file serves with 503 under low heap and rate-limits them while a WS stream is active so file-serving TX can't starve the 10 Hz broadcast.
- **Power** — keep the scale awake while a WiFi client is connected (the 15-min auto-off was reset only by BLE/button, so a battery-powered WiFi-only client died mid-stream).
- **Unified 10 Hz tick** — one grid-advanced 100 ms timer drives BLE/USB/WS each at its own multiple, replacing four drift-prone per-interface timers; true ~10 Hz (was ~9.6 Hz under load). Per-interface rates/packets unchanged.
- **Test tooling** — stress + rate-measurement harnesses in `tools/`.

**Root cause of the reported drops:** under BT/WiFi coexistence, transient radio stalls left WS data unacked past AsyncTCP's default 5 s ACK timeout, so the firmware closed the socket (graceful FIN, no WS close frame). A separate connection-churn path could exhaust heap and wedge the IP stack while the main loop kept running. (A misbehaving AP was also a major field factor, since fixed.)

## Test plan / validation (on hardware, ESP32-S3)
- [x] >1 hr continuous WiFi 10 Hz stream — **0 drops**, single connection, under WS/HTTP churn + mDNS load
- [x] 30-min concurrent **WiFi + USB** both at true **10.0 Hz** (18,030 USB frames, 0 corrupt; 0 WiFi drops/stalls)
- [x] Source cadence verified — scale emits every ~100 ms (p50 = 100, max ≤ 112 ms)
- [x] Broadcast priority — under an HTTP flood that previously starved a client to 0 Hz, now ~9.9 Hz
- [x] BLE confirmed live + smooth in the Decenza app throughout (BLE code byte-unchanged)
- [ ] Reviewer: confirm on your hardware/AP

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,12 +24,15 @@ After this, one can simply use `$pio run -t upload` and platformio will build an
 
 ## Wifi mode
 
-Decent Scale now uses WiFi for additional features, not just OTA update.   
+Decent Scale uses WiFi for web apps, WebSocket data streaming, and OTA updates.
 
 To enable WiFi mode, go to HDS setup menu and find "Wifi settings" entry. From there you can enable/disable WiFi as well as see current WiFi details. If you toggle WiFi on/off, a restart of the scale is required for the new settings to take effect.
-Initially HDS will open its own WiFi, called "Decent Scale", protected with a password '12345678'. Once connected to this WiFi, you can navigate with your browser to [hds.local](http://hds.local) or [192.168.1.1](http://192.168.1.1) to change the Wifi settings, in case you want HDS to connect to your home WiFi. 
-Again, make sure to restart the scale for the settings to take effect.
-If you store WiFi settings incorrectly, or you take your HDS out of signal range, HDS will return back to its own Wifi (Decent Scale) - so you can change the settings again if needed.
+
+**First-time setup:** if no WiFi credentials are stored, the scale opens its own access point — SSID `Decent Scale`, password `12345678`, IP `192.168.1.1`. Connect to it and navigate to [hds.local](http://hds.local) to enter your home WiFi credentials. The scale will restart and connect to your network.
+
+**Reconnect:** once configured, the scale stays in STA mode and reconnects automatically after signal loss with exponential backoff (5 s → 10 → 20 → 40 → 60 s cap). It no longer falls back to the AP — a configured scale keeps retrying your home network.
+
+**Discovery:** the scale advertises itself via mDNS (`hds.local`) and DNS-SD (`_decentscale._tcp`) so apps can discover it on the LAN without knowing the IP.
 
 ## Web apps
 The same web apps that could be used with Half Decent Scale from [the web](https://decentespresso.com/docs/introducing_half_decent_scale_web_apps) have now been rewritten to run directly from the Half Decent Scale.
@@ -128,12 +131,23 @@ info
 The legacy text `tare` command is intentionally silent for backwards
 compatibility. JSON `{ "command": "tare" }` returns a `type: "status"` ack.
 
-The same commands can be sent as JSON:
+The same commands can be sent as JSON (full form):
 
 ```json
 { "command": "tare" }
 { "command": "timer", "action": "start" }
 ```
+
+**JSON shorthand** — control keys at the top level are also accepted:
+
+```json
+{ "events": "on" }
+{ "display": "off" }
+{ "tare": true }
+{ "timer": "start" }
+```
+
+Bool values map to on/off (`{"display":false}` → off). The `power` key is excluded from the bool form — power-off requires the explicit `{"command":"power","action":"off"}` or text `power off`.
 
 An unrecognized or malformed command (bad JSON, or an unknown verb) returns an
 `error` frame rather than silence, so a client can distinguish a rejected
--- a/include/ble.h
+++ b/include/ble.h
@@ -629,13 +629,12 @@ void buildWeightPacket(byte data[7]) {
   data[6] = calculateXOR(data, 6);
 }
 
+// Rate-gating is handled centrally by the unified weight-output tick in the
+// main loop (which honors weightBleNotifyInterval, i.e. the per-client 2k/5k/10k
+// rate). This just emits one notification when called; the active-state check
+// stays so we never touch the characteristic without a live connection.
 void sendBleWeight() {
   if (!(b_ble_enabled && deviceConnected && pReadCharacteristic)) return;
-
-  unsigned long currentMillis = millis();
-  if (currentMillis - lastBleWeightNotifyTime < weightBleNotifyInterval) return;
-  lastBleWeightNotifyTime = currentMillis;
-
   byte data[7];
   buildWeightPacket(data);
   pReadCharacteristic->setValue(data, 7);
--- a/include/ble.h
+++ b/include/ble.h
@@ -630,8 +630,8 @@ void buildWeightPacket(byte data[7]) {
 }
 
 // Rate-gating is handled centrally by the unified weight-output tick in the
-// main loop (which honors weightBleNotifyInterval, i.e. the per-client 2k/5k/10k
-// rate). This just emits one notification when called; the active-state check
+// main loop (which honors weightBleNotifyInterval, fixed at 100 ms / 10 Hz for
+// BLE). This just emits one notification when called; the active-state check
 // stays so we never touch the characteristic without a live connection.
 void sendBleWeight() {
   if (!(b_ble_enabled && deviceConnected && pReadCharacteristic)) return;
--- a/include/declare.h
+++ b/include/declare.h
@@ -24,7 +24,7 @@ BLEService *pService = NULL;
 BLEAdvertising *pAdvertising = NULL;
 BLECharacteristic *pReadCharacteristic = NULL;
 BLECharacteristic *pWriteCharacteristic = NULL;
-bool deviceConnected = false;
+volatile bool deviceConnected = false;  // written from the BLE task, read on the main loop (heap watchdog)
 
 // The model byte is always 03 for Decent scales
 const byte modelByte = 0x03;
--- a/include/declare.h
+++ b/include/declare.h
@@ -24,7 +24,10 @@ BLEService *pService = NULL;
 BLEAdvertising *pAdvertising = NULL;
 BLECharacteristic *pReadCharacteristic = NULL;
 BLECharacteristic *pWriteCharacteristic = NULL;
-volatile bool deviceConnected = false;  // written from the BLE task, read on the main loop (heap watchdog)
+// volatile: written from the BLE (NimBLE) task in onConnect/onDisconnect; read
+// on the main loop (power-off keepalive in loop(), heap-watchdog gate in
+// wifiSupervise()). volatile prevents the readers caching a stale value.
+volatile bool deviceConnected = false;
 
 // The model byte is always 03 for Decent scales
 const byte modelByte = 0x03;
--- a/include/parameter.h
+++ b/include/parameter.h
@@ -9,9 +9,13 @@ unsigned long lastBleWeightNotifyTime = 0;  // Stores the last time the weight n
 unsigned long lastUsbWeightNotifyTime = 0;  // Stores the last time the weight notification was sent
 
 unsigned long lastWeightTextNotifyTime = 0;   // Stores the last time the weight notification was sent
-unsigned long weightBleNotifyInterval = 100;  // Interval at which to send weight notifications (milliseconds)
-unsigned long weightUsbNotifyInterval = 100;  // Interval at which to send weight notifications (milliseconds)
-unsigned long weightTextNotifyInterval = 1000;
+unsigned long weightBleNotifyInterval = 100;  // BLE notify interval (ms); per-client (2k/5k/10k => 500/200/100)
+unsigned long weightUsbNotifyInterval = 100;  // USB binary notify interval (ms)
+unsigned long weightTextNotifyInterval = 1000;  // USB text/debug line interval (ms)
+// Base period of the unified weight-output tick. One grid timer drives every
+// interface; each sends every (its NotifyInterval / base) ticks. All the
+// per-interface intervals above (and the WS ones below) are multiples of this.
+const unsigned long WEIGHT_BASE_INTERVAL_MS = 100;
 const unsigned long WEBSOCKET_2HZ_NOTIFY_INTERVAL_MS = 500;
 const unsigned long WEBSOCKET_5HZ_NOTIFY_INTERVAL_MS = 200;
 const unsigned long WEBSOCKET_10HZ_NOTIFY_INTERVAL_MS = 100;
--- a/include/parameter.h
+++ b/include/parameter.h
@@ -5,11 +5,7 @@
 //ble
 bool b_ble_enabled = false;
 bool b_usbweight_enabled = false;
-unsigned long lastBleWeightNotifyTime = 0;  // Stores the last time the weight notification was sent
-unsigned long lastUsbWeightNotifyTime = 0;  // Stores the last time the weight notification was sent
-
-unsigned long lastWeightTextNotifyTime = 0;   // Stores the last time the weight notification was sent
-unsigned long weightBleNotifyInterval = 100;  // BLE notify interval (ms); per-client (2k/5k/10k => 500/200/100)
+unsigned long weightBleNotifyInterval = 100;  // BLE notify interval (ms). Fixed at 100ms (10Hz); not runtime-configurable over BLE.
 unsigned long weightUsbNotifyInterval = 100;  // USB binary notify interval (ms)
 unsigned long weightTextNotifyInterval = 1000;  // USB text/debug line interval (ms)
 // Base period of the unified weight-output tick. One grid timer drives every
--- a/include/usbcomm.h
+++ b/include/usbcomm.h
@@ -582,7 +582,7 @@ void sendUsbWeight() {
 // main loop; this just emits one line when called.
 void sendUsbTextWeight() {
   Serial.print(millis());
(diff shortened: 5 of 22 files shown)


skialpine:
### Code review

Found 8 issues. The reliability work looks solid and was hardware-validated; these are correctness/cleanup items found by review.

**1. mDNS is torn down and re-advertised from the WiFi event task on every reconnect, and double-registered on boot**

`onWifiEvent()` runs in the async WiFi event task and calls `MDNS.end(); setupMdns();` on `GOT_IP`. `setupMdns()` (which does `MDNS.begin()` + `addService`) is *also* called unconditionally in `setupWifi()`, so on a normal boot the service registers twice (ESP32 `addService` doesn't dedupe), and the teardown/rebuild reruns from event-task context on every reconnect — the exact churn this PR targets. Consider deferring it to `wifiSupervise()` (main loop) and dropping the redundant boot call.

https://github.com/decentespresso/openscale/blob/0ed55903b569d58799b24d1fb6f8b63f71c0ce2c/src/wifi_setup.cpp#L133-L136

**2. WiFi credentials may not persist before `esp_restart()`, and AP fallback was removed**

`saveCredentials()` writes the SSID/pass to NVS and the web handler calls `esp_restart()` immediately after; NVS commits are lazy, so the restart can truncate the write. Because this PR removed the AP-mode fallback from `connectToWifi()`, a scale that fails to persist credentials now has no recovery path and retries forever against nothing. An explicit `preferences.end()` (or short delay) before restart would close the gap.

https://github.com/decentespresso/openscale/blob/0ed55903b569d58799b24d1fb6f8b63f71c0ce2c/src/wifi_setup.cpp#L268-L271

https://github.com/decentespresso/openscale/blob/0ed55903b569d58799b24d1fb6f8b63f71c0ce2c/include/webserver.h#L41-L44

**3. Heap-watchdog "sustained 2 s" guard is defeated by a stale `lowHeapSince` after partial recovery**

`lowHeapSince` is only cleared when free heap rises above `HEAP_CRITICAL + 5000` (20000). If heap dips below 15000 (setting the timer), recovers to ~18000 and hovers there, the timer is never reset; a later brief dip below 15000 then satisfies `now - lowHeapSince >= 2000` on the *first* sample and reboots immediately, instead of requiring a fresh 2 s window. Clear `lowHeapSince` whenever `freeHeap >= HEAP_CRITICAL`.

https://github.com/decentespresso/openscale/blob/0ed55903b569d58799b24d1fb6f8b63f71c0ce2c/src/wifi_setup.cpp#L201-L213

**4. HTTP admission control 503s most web-app assets while a WS stream is active**

While `websocket.count() > 0`, the middleware allows one non-`/snapshot` request per 200 ms via a single shared `lastServe`. The on-device apps served from LittleFS pull multiple local assets (e.g. `dosing_assistant` loads a CSS file plus an ES-module graph under `modules/`), and browsers do not retry 503 for sub-resources — so loading/reloading an app while any client is streaming (the apps hold a persistent WS) will 503 most of its assets and render it broken. The comment's "clients retry" assumption doesn't hold for static sub-resources. Consider exempting same-origin static assets or using a small token bucket.

https://github.com/decentespresso/openscale/blob/0ed55903b569d58799b24d1fb6f8b63f71c0ce2c/include/webserver.h#L82-L90

**5. `b_wifiEnabled` is toggled as an LED-blink flag during association but now gates real behavior**

`connectToWifi()` toggles `b_wifiEnabled` ~1 Hz to blink during association, but the loop now uses it as a real gate for the power-off keepalive and the WiFi housekeeping/weight-tick blocks. During association it's `false` ~50% of the time, so those paths intermittently treat WiFi as off. Benign at boot, but a latent hazard now that the flag carries control-flow meaning — a dedicated `b_wifiAssociating` blink flag would remove the ambiguity.

https://github.com/decentespresso/openscale/blob/0ed55903b569d58799b24d1fb6f8b63f71c0ce2c/src/wifi_setup.cpp#L70-L72

https://github.com/decentespresso/openscale/blob/0ed55903b569d58799b24d1fb6f8b63f71c0ce2c/src/hds.ino#L1291-L1293

**6. `connectToWifi()` background-retry path logs "Connected to" / `0.0.0.0` and sets `b_wifiEnabled = true` while not connected**

When the connect loop times out and breaks to "retry in background", it falls through unconditionally to the success prints (`Serial.println("Connected to ")`, `WiFi.localIP()` → `0.0.0.0`) and `b_wifiEnabled = true`, regardless of `WiFi.status()`. The supervisor still recovers, but the log is misleading and the flag is set true on the not-connected path. Guard lines 83-87 with `if (WiFi.status() == WL_CONNECTED)`.

https://github.com/decentespresso/openscale/blob/0ed55903b569d58799b24d1fb6f8b63f71c0ce2c/src/wifi_setup.cpp#L82-L87

**7. Dead code: orphaned weight-timer globals and the never-read `g_staIntended`**

The unified-tick refactor removed the per-function rate gates, leaving `lastBleWeightNotifyTime` / `lastUsbWeightNotifyTime` / `lastWeightTextNotifyTime` with no readers or writers (they're non-`static` globals instantiated in every TU that includes the header). Separately, `g_staIntended` is written in `connectToWifi()`/`setupAP()` but never read — the supervisor actually gates on `params.hasCredentials()`, so the flag (and its comment) misleads about the real control path. Remove both.

https://github.com/decentespresso/openscale/blob/0ed55903b569d58799b24d1fb6f8b63f71c0ce2c/include/parameter.h#L8-L11

https://github.com/decentespresso/openscale/blob/0ed55903b569d58799b24d1fb6f8b63f71c0ce2c/src/wifi_setup.cpp#L16-L18

**8. Comments describe behavior the code doesn't have**

The "per-client 2k/5k/10k" rate language for BLE is inaccurate: `weightBleNotifyInterval` is hardcoded to 100 ms with no BLE command that changes it; the 2k/5k/10k labels are WebSocket-only (parameter.h:12, ble.h:632-633, hds.ino:1392). Separately, the `wifiSupervise()` header comment omits that the function can `esp_restart()` the device via the heap watchdog — a significant undocumented side effect.

https://github.com/decentespresso/openscale/blob/0ed55903b569d58799b24d1fb6f8b63f71c0ce2c/include/ble.h#L631-L634

https://github.com/decentespresso/openscale/blob/0ed55903b569d58799b24d1fb6f8b63f71c0ce2c/include/wifi_setup.h#L11-L14

<details>
<summary>5 below-threshold issues (50-74)</summary>

- [Worth a look] Heap watchdog calls `esp_restart()` with no `deviceConnected` gate, so a heap storm can reboot mid-shot and drop a live BLE session — against the "BLE must not get less reliable" goal (score: 59)
- [Worth a look] `HTTP_MIN_INTERVAL_MS` in the webserver.h:70 comment doesn't match the real constant `HTTP_MIN_INTERVAL_WHILE_STREAMING_MS` (score: 50)
- [OK to skip] Unified-tick comment says "Replaces four independent timers" but only three rate gates existed before (USB binary was previously ungated) (score: 50)
- [OK to skip] Stale "7 bytes of data" comment in `sendUsbTextWeight()` — it emits variable-length ASCII, not a 7-byte frame (score: 50)
- [OK to skip] WiFi reconnect (`WiFi.disconnect()`/`begin()` from the main loop) isn't BLE-coexistence-aware, but it's a documented, backoff-mitigated tradeoff and a net improvement over the old AP-fallback behavior (score: 50)

</details>

skialpine:
### Review fixes pushed

Addressed the review findings in 4 focused commits (`fee6446`, `465dbc7`, `e90746a`, `3a67ad6`). Built for `esp32s3` and flashed to hardware; re-verified on-device.

**Fixed**
1. **mDNS** — `GOT_IP` now sets a pending flag and `wifiSupervise()` does the `MDNS.end()/setupMdns()` on the main loop, off the WiFi-event task. The redundant boot-time `setupMdns()` is gone for the STA path (kept for AP, which emits no `GOT_IP`), so the service no longer double-registers.
3. **Heap watchdog stale timer** — `lowHeapSince` is now cleared whenever heap recovers above critical, removing the 15–20 KB dead band that could trigger an instant reboot on a later brief dip.
4. **HTTP 503 vs web apps** — replaced the single 1-per-200ms gate with a token bucket (depth 12) so a full page-load asset graph bursts through while streaming, then sustained floods are throttled at the refill rate.
5. **`b_wifiEnabled` blink flag** — no longer toggled during association; the OLED icon blink is now driven by a time+status check, so the flag stays monotonic for the control gates.
6. **False "Connected" log** — the success log/IP print is now guarded by `WL_CONNECTED`; the background-retry path no longer prints `0.0.0.0`.
7. **Dead code** — removed the three orphaned `lastXxxWeightNotifyTime` globals and the never-read `g_staIntended`.
8. **Comments** — corrected the BLE "per-client 2k/5k/10k" claims (BLE is fixed 100 ms/10 Hz), documented the `wifiSupervise()` heap-watchdog reboot, fixed the `HTTP_MIN_INTERVAL_MS` name and the "four timers"→three / "7 bytes" comments.

Also from the below-threshold list: the **heap watchdog now skips `esp_restart()` while a BLE client is connected** (defers until BLE disconnects) so a WiFi-side OOM can't kill a live shot.

**Intentionally not changed**
- **#2 (NVS commit race)** — false positive: arduino-esp32 `Preferences::putString()` calls `nvs_commit()` synchronously, so credentials persist before `esp_restart()`. The AP-fallback removal is the documented intent of this PR, so there's no recovery-path regression.
- **Reconnect coexistence** — the `WiFi.disconnect()/begin()` reconnect from the main loop is a documented, backoff-mitigated tradeoff and a net improvement over the old AP-fallback-and-die behavior.

**On-device re-verification**
- WS weight stream: clean **10.00 Hz** (scale emission p50 = 100 ms, 0 gaps > 150 ms)
- `hds.local` resolves after reboot (deferred mDNS advertise works)
- All 12 web-app assets serve `200` (0× 503) while a WS client is streaming (token bucket)
- BLE was not independently checked from the host (can't observe BLE here); its send path is behavior-unchanged and the watchdog change only adds protection.

skialpine:
### Code review (re-review after fix commits)

Re-reviewed the 4 fix commits (`fee6446`, `465dbc7`, `3a67ad6`, `e90746a`). All 8 findings from the previous review are genuinely resolved, and the #2 "false positive" call (NVS `putString` commits synchronously) is confirmed correct. The fixes did introduce 2 small new issues, both in the heap-watchdog area.

Found 2 issues:

1. The new BLE-defer branch logs on every main-loop pass with no rate limit. Once heap is critical for >2 s while a BLE client is connected, `lowHeapSince` is never reset, so this `Serial.printf` fires every loop iteration (the loop has no sleep), flooding Serial — unlike the `[health]` log right below it, which is gated by a 5 s `lastLog`. Add a `lastDeferLog` guard.

https://github.com/decentespresso/openscale/blob/e90746a8dacd9da3fe55df37059ed0d2a68a8f94/src/wifi_setup.cpp#L218-L226

2. `deviceConnected` is now read cross-task in the watchdog but is a plain `bool`. It's written from the BLE (NimBLE) task and read on the main loop; every other cross-task flag in this PR (`g_mdnsAdvertisePending`, `g_wifiInitDone`, `g_wifiDisconnects`/`Reconnects`) is `volatile`. Mark it `volatile` in `declare.h` for consistency and to prevent the compiler caching a stale value in the watchdog check.

https://github.com/decentespresso/openscale/blob/e90746a8dacd9da3fe55df37059ed0d2a68a8f94/include/declare.h#L26-L28

<details>
<summary>3 below-threshold issues (50-74)</summary>

- [Worth a look] `g_mdnsAdvertisePending` is a single bool with a non-atomic read-clear; a `GOT_IP` that fires while `setupMdns()` is mid-run can be swallowed, so after a flapping-link reconnect `hds.local` may stop being advertised until the next clean reconnect (discovery-only impact). Re-arm/re-check the flag after `setupMdns()`. (score: 68) — `src/wifi_setup.cpp` L196-L201
- [Worth a look] Heap watchdog defers `esp_restart()` indefinitely while a BLE client stays connected, so a real lwIP wedge under WiFi/HTTP churn during an always-connected BLE session has no auto-recovery until BLE drops. This is an intentional tradeoff (never reboot mid-shot), but bounding the deferral (long window or hard heap floor) would still self-heal a true wedge. (score: 62) — `src/wifi_setup.cpp` L213-L231
- [OK to skip] Unified-tick comment says "three independent timers"; accurate, but a prior commit message said "four", so a reader diffing history may be briefly confused. (score: 50)

</details>

skialpine:
### Re-review fixes + 30-min regression soak

Pushed 2 follow-up commits (`226582e`, `5af1568`) addressing the re-review, then re-flashed and ran a fresh 30-minute concurrent soak.

**Re-review follow-ups (`226582e`)**
- Heap-watchdog BLE deferral is now bounded to 60 s (`HEAP_CRITICAL_BLE_DEFER_MAX`): a genuinely wedged stack still self-heals even if a BLE client stays connected, but a normal sub-minute shot is never interrupted.
- The "defer reboot" log is rate-limited to once / 5 s (it previously printed every loop pass once the window expired).
- `deviceConnected` is now `volatile` (written from the BLE task, read in the watchdog on the main loop), consistent with the other cross-task flags.
- Documented that clearing `g_mdnsAdvertisePending` before `setupMdns()` re-advertises a reconnect that arrives mid-rebuild on the next pass (so it's not lost).

**30-min soak results** (WS 10 Hz + USB 10 Hz + HTTP/WS churn + mDNS, concurrent):
- WS stream: **0 drops** over 1763 s, single connection held throughout.
- USB: 17,834 frames, **0 bad XOR, ~10 Hz** steady (the one large inter-frame gap is the USB-open reboot at t=0).
- Connection churn: **879 connects (440 HTTP / 439 WS), 0 fail, 0 hang** — no server stalls.
- mDNS: resolver **339/339 (100%)**, 1% UDP query loss — the deferred main-loop advertise works.

**One honest correction on finding #4 (web apps while streaming):** my earlier "token bucket lets a full asset graph burst through" was overstated. The token bucket does fix the *rate-limit* dimension (sequential/staged loads pass freely), but a truly **concurrent** 12-asset burst while a client streams is shed by the *pre-existing* memory-pressure layer (`free heap < 40000` → "low memory, retry"), because each static serve costs ~25 KB transient on a ~70 KB-heap device. That shedding is intentional OOM protection — it's exactly why this soak saw 0 hangs/wedges under churn — so I did not loosen it. Net: single-client page loads and staged/sequential loads work; heavy concurrent module-graph loads while a second client streams will still 503 some sub-resources. A proper fix would be client-side (staged/retry loading) or a serialized static-serve queue; flagging rather than trading away the OOM safety.

skialpine:
### Code review

Reviewed the commits since the last review (`e90746a..553e70a`). All findings from the previous two rounds are confirmed resolved, and the deep bug scan found no high-severity bugs introduced. One minor finding:

1. The new JSON control shorthand handles `{"key":"off"}` but not `{"key":false}` for commands with an off-state. The bool branch only dispatches on `true` (`doc[key].as<bool>()`), so `{"sleep":false}` / `{"display":false}` / `{"low_power":false}` fall through to an `unknown_command` error, while `{"sleep":"off"}` works. Inconsistent for a newly-added convenience shorthand (no current client sends the bool-false form, so impact is low). Either map `false` to the off action for those keys, or document that the string form is canonical.

https://github.com/decentespresso/openscale/blob/553e70a0b414227091c0c4c427fbb58a6a4190b0/include/websocket.h#L505-L514

<details>
<summary>3 below-threshold issues (50-74)</summary>

- [Worth a look] Heap watchdog: `lastDeferLog` isn't reset on heap recovery, so the first "defer reboot" log of a subsequent low-heap episode within 5 s can be suppressed. Logging-only; one-line fix (`lastDeferLog = 0;` beside `lowHeapSince = 0;`). (score: 50)
- [OK to skip] `setupMdns()` was changed to return `bool` but both call sites ignore it; the function already logs `MDNS.begin` failure internally, so observability is unaffected. (score: 50)
- [OK to skip] `declare.h` comment attributes `deviceConnected` writes to the "BLE (NimBLE) task" — accurate for the esp32s3 build (NimBLE under `CONFIG_NIMBLE_ENABLED`), just could be phrased more explicitly. (score: 50)

</details>

tadelv (approved):
Comprehensive quality-of-life upgrade. Root cause analysis is solid (AsyncTCP ACK timeout + no reconnect recovery + drift-prone timers), and the fixes are well-designed with careful attention to threading safety.

Highlights:
- Reconnect supervisor with exponential backoff — proper quiet gaps to avoid AP rate-limiting
- Heap watchdog with BLE-deferral cap (60s) — protects shots while still self-healing
- Unified 10 Hz grid-advanced tick — eliminates a real drift-bug class
- Token-bucket HTTP admission control — lets page loads burst through while protecting the stream
- mDNS deferral to main loop — avoids cross-task races
- 11 test harnesses with real hardware validation (>1 hr WiFi with 0 drops, 30-min concurrent WiFi+USB at true 10.0 Hz)

Added a commit to update the README: removed the stale AP-fallback claim and documented the new JSON command shorthand forms.

Well done.

PR #57, opened 2026-05-25
Scale telemetry: SoC temp, weight-stall watchdog, reset reason

skialpine:
## Summary
Diagnostics for field reports of "weight stops being collected under sustained multi-protocol load" — the only recovery seen was a long battery-out cooldown (a quick power-cycle didn't help), pointing at a thermal/analog failure rather than firmware state. Adds telemetry to confirm/rule it out, with **no behavior change** to the weight/WiFi/BLE paths.

New fields in the `/snapshot` WS status frame (and serial logs):
- `soc_temp_c` / `soc_temp_max_c` — live + peak ESP32-S3 die temp (`temperatureRead()`), sampled every 2 s.
- `weight_stalled` / `stall_count` / `last_stall_ms` / `last_stall_temp_c` — a watchdog in `pureScale()` that flags when the ADS1232 raw value is frozen/railed >8 s (a live cell dithers every sample), recording the die temp at failure to correlate stalls with heat. Skipped during the deliberate ADC power-cycle recovery; throttled to 250 ms.
- `reset_reason` — `esp_reset_reason()` at boot, so a brownout/panic/WDT reset is attributable.

Plus `tools/thermal_load_test.sh`: a 1-hour USB+WiFi+churn+mDNS soak that polls the telemetry (BT driven externally).

Threading: new cross-task scalars are `volatile` per CLAUDE.md; the status frame only reads them.

## Test plan
- [x] Builds for esp32s3; flashed; status frame reports the new fields.
- [ ] 1-hour multi-protocol soak to capture peak temp + any stall and the die temp at which it occurs.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -158,6 +158,10 @@ This document is meant to evolve with the codebase. During a session, if you (Cl
 
 If you fix a bug whose symptom is documented in the "When something is broken" table, leave the entry in place — it's still the right "first place to look" for the next person.
 
+## Fixing bugs you find along the way
+
+Pre-existing bugs get fixed too — "it was already there" is not a reason to defer. When you turn up a bug while working on something else (a review flags it, you read past it, a test surfaces it), fix it as part of the same change; a pre-existing bug is no less bad than a newly introduced one, and the person touching the code is the right person to fix it. The only exception is when the fix is genuinely a large, independent effort — then call it out explicitly and agree on a separate change, rather than silently leaving it in place.
+
 ## Don't
 
 - Don't call I²C / SPI / blocking IO from the AsyncTCP task.
--- a/README.md
+++ b/README.md
@@ -162,14 +162,12 @@ command from a frame that never arrived:
 }
 ```
 
-Status frame shape:
+Status frame shape (lean — only fields that change during a session):
 
 ```json
 {
   "type": "status",
   "status": "ok",
-  "protocol_version": 1,
-  "firmware_version": "FW: 3.0.9",
   "grams": 25.66,
   "ms": 12345,
   "battery_percent": 82,
@@ -189,8 +187,10 @@ Status frame shape:
 ### `session_info` frame (per-connect, server → client)
 
 Sent once to each client immediately after WebSocket handshake. Carries the
-fields that don't change for the life of the connection, so the hot status
-broadcast doesn't have to ship them on every tick.
+fields that don't change for the life of the connection (protocol version,
+firmware version, reset reason), so the hot status broadcast doesn't have to
+ship them on every tick. Clients can also re-request it explicitly by sending
+`{"command":"session_info"}` (alias: `session`).
 
 ```json
 {
--- a/README.md
+++ b/README.md
@@ -186,6 +186,81 @@ Status frame shape:
 }
 ```
 
+### `session_info` frame (per-connect, server → client)
+
+Sent once to each client immediately after WebSocket handshake. Carries the
+fields that don't change for the life of the connection, so the hot status
+broadcast doesn't have to ship them on every tick.
+
+```json
+{
+  "type": "session_info",
+  "protocol_version": 1,
+  "firmware_version": "FW: 3.0.9",
+  "reset_reason": "poweron",
+  "ms": 12345
+}
+```
+
+### `debug` frames (diagnostic telemetry, opt-in)
+
+Diagnostic telemetry — SoC die temperature, weight-stall watchdog state, ADC
+power-cycle recovery count — is delivered out-of-band from the hot status
+broadcast so apps that don't care about it (the on-device UI, the
+decentespresso app, third-party scale apps) aren't paying ~21% extra bytes per
+status tick.
+
+**On-request snapshot** — send `{"command":"debug"}` (also accepted: `diag`)
+to get the full diagnostic set per-client. Reply:
+
+```json
+{
+  "type": "debug",
+  "status": "ok",
+  "soc_temp_c": 33.3,
+  "soc_temp_max_c": 41.2,
+  "weight_stalled": false,
+  "stall_count": 0,
+  "last_stall_ms": 0,
+  "last_stall_temp_c": 0.0,
+  "adc_recovery_count": 0,
+  "ms": 12345
+}
+```
+
+**Event broadcasts** — emitted when something diagnostic-relevant changes.
+Subscribers keep their own snapshot from `session_info` + the on-request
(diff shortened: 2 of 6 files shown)


skialpine:
### Code review

Reviewed the telemetry diff (deep bug scan + threading/CLAUDE.md check). Found one real issue, now fixed in this PR (commit `8e345d9`):

- **Stall watchdog false-trip during ADC recovery** — the watchdog read `getDebugInfo().rawValue`, which is frozen by definition during the firmware's own `powerDown()/powerUp()` recovery, so a genuine signal-timeout recovery would be miscounted as a railed/frozen stall (corrupting the very metric this adds). Fixed: skip the check while `b_adc_recovery_active` and re-seed the window on resume.
- Also (cost): it ran `getDebugInfo()` (which does a `sqrt` + dataset passes) every loop iteration though only `rawValue` is used — wasteful on a chip we're characterizing for heat. Now throttled to 250 ms (the ADC only samples ~10/s).

Verified clean: printf format/arg pairing in both status frames; cross-task reads are `volatile` (benign torn read only); `temperatureRead()`/`resetReasonStr()` safe; at-rest false-positive risk is low (24-bit raw at SAMPLES=1 dithers every sample, so 8 s of byte-identical raw is a genuine freeze).

skialpine:
Force-pushed: rebased onto current main (post #58 + #60) and restructured telemetry delivery.

**What changed in the protocol:** the 8 telemetry fields no longer ride on the hot `status` broadcast. They move out into three new frame shapes:

- **`session_info`** — sent once per client on WS connect. Carries `protocol_version`, `firmware_version`, `reset_reason` (the connection-immutable fields).
- **`debug` events** — broadcast on actual diagnostic changes: `stall_start` / `stall_end`, `adc_recovery` (count incremented), `temp_peak` (new SoC max). Subscribers get the event the moment it happens; non-subscribers ignore the unknown type.
- **`debug` reply** — per-client on-request snapshot via `{\"command\":\"debug\"}`.

Status frame stays at its current 16 fields (~310 B payload) instead of bloating to ~470 B. Apps that don't care about diagnostics get back ~21% per broadcast.

**Why now:** the 8-hour soak validating #61's heap-defense knobs (cap=4 + 32K floor) showed that under marginal RSSI (-83) and multi-client load, status broadcasts were the dominant heap-pressure source — gate fired 5,000+ times. Slimming the status payload at the source is structurally better than just gating harder.

**Backward compat:** the previous `status` frame shape (which had the 8 telemetry fields appended) had not shipped on any merged release, so there's nothing to deprecate. Clients written against the README's old example will fall back to ignoring unknown fields cleanly anyway.

Build clean, validated diff against current main. Squashed to one commit since the original 7-commit history was reshuffled enough to make rebase-without-conflict impossible — happy to break it back into reviewable chunks if that's preferred.

skialpine:
### Code review

Found 3 issues:

1. `tools/thermal_load_test.sh` reads telemetry fields exclusively from `status` frames, but commit `f72d6e2` moved them all to `session_info` and `debug` frames. The monitor never subscribes to `debug` events nor reads `session_info`, so `peak` stays at its `-999` sentinel and the verdict path forces `RESULT=FAIL` on every run regardless of device behavior. Fix: handle all three frame types in the Python monitor (the script also unconditionally sends `events on` but never `debug` subscribe). The script's header comment at line 5 still says "WS status-frame telemetry" — also stale.

https://github.com/decentespresso/openscale/blob/f72d6e2adc7b7ebff7f1f2a4bb2e5fcc7e76f961/tools/thermal_load_test.sh#L67-L96

2. Debug event broadcasts (`sendWebsocketDebugStall`, `sendWebsocketDebugAdcRecovery`, `sendWebsocketDebugTempPeak`) gate only on `!b_wifiEnabled || websocket.count() == 0`. They skip the `b_websocketEventsEnabled` check that every other event broadcast in this file uses (`sendWebsocketButton` L188, `sendWebsocketPowerOff` L199, `sendWebsocketStatusAll` L246). README says "A client must send `events on` before periodic status, local scale button presses, or power-off notifications are emitted." — new debug frames break this opt-in contract. Either add `!b_websocketEventsEnabled` to the early-return, or document the always-on choice explicitly in the README's debug-frame section.

https://github.com/decentespresso/openscale/blob/f72d6e2adc7b7ebff7f1f2a4bb2e5fcc7e76f961/include/websocket.h#L317-L350

3. `i_adc_recovery_count` documentation contradicts the implementation. `include/parameter.h:197-200` says "counts truthfully ... over a long soak instead of saturating at 255" and README:256 says "number of ADC power-cycle recoveries since boot". But `src/hds.ino:1011` calls `resetAdcRecoveryState()` after every successful `scale.update()`, which zeroes the counter. The actual semantics — confirmed by the display-logic use at `src/hds.ino:1830` (`>= ADC_ERROR_RECOVERY_COUNT ? "ADC ERROR" : "ADC RECOVER"`) — is "consecutive failed recoveries in the current episode," not a lifetime total. A client tracking adc_recovery_count expecting monotonic growth will see it ping-pong to 1 and back to 0. Either fix the docs (parameter.h block-comment + README field description) or rename for a true lifetime counter.

https://github.com/decentespresso/openscale/blob/f72d6e2adc7b7ebff7f1f2a4bb2e5fcc7e76f961/include/parameter.h#L195-L201

---

<details>
<summary>1 below-threshold issue (50-74)</summary>

- [Worth a look] `src/hds.ino:22` comment says `adsDebugCallback` is "enabled during soak tests" — the actual enablement is via the USB binary protocol in `include/usbcomm.h`, not anything soak-test-specific. One-line clarification (score: 50).

</details>

tadelv:
Closing in favor of a smaller, focused follow-up that extends the existing ADS1232 debug packet (USB binary 0x25 + BLE) with reset_reason only. Dropping: SoC temperature sampling (thermal ruled out as the culprit), weight-stall watchdog, and adc_recovery_count exposure (software ADC reset via PWR_CTRL toggle has not been shown to help on PCBs <= 8.3.1, so the counter doesn't earn a protocol field).

tadelv:
One more note on direction for anyone returning to this: if/when there's appetite to revive a WiFi-side telemetry channel for these fields, the preferred path is to extend the existing ADS1232 debug packet first (USB binary 0x25 + BLE notify on fff4) — i.e. follow the model #63 establishes for reset_reason at byte 24 — and only then bridge it onto WS. That keeps a single source-of-truth packet across all three transports instead of bifurcating into a WiFi-specific status/session/debug shape that USB and BLE can't see. The bytes 25-37 region of the 41-byte packet still has room for additional fields, and any new WS layer can subscribe to the same bytes rather than redefining them.

Issue #23, opened 2025-10-23
Currently active settings display


tadelv:
Have a place, (preferably) both on HDS display and web interface, of the currently active settings. 
Examples:
```
WiFi: enabled, SSID saved
BLE Buttons active: true
BLE Heartbeat: enabled
Auto Sleep: enabled (after 15minutes)
```

tadelv:
closed with #64

PR #64, opened 2026-06-01
Add active settings status display

ODevStudio:
Refs #23

Adds a current settings display to:
- OLED setup menu via a new top-level Status item
- root web UI via the existing /snapshot WebSocket status frame

Also extends the WebSocket status frame with WiFi, BLE, heartbeat, and auto sleep settings.

Tested:
- pio run -e esp32s3
- flashed firmware and LittleFS to ESP32-S3
- HDS Settings page
- Wifi web page

### DIFF
--- a/include/menu.h
+++ b/include/menu.h
@@ -394,7 +394,7 @@ void flipScreenOn() {
   actionMessage = "Flip On";
   t_actionMessage = millis();
   t_actionMessageDelay = 1000;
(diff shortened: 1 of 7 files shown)


tadelv:
Thanks @ODevStudio , this looks really neat.
The initial list of settings in #23 is a start, could you also ask the quick boot and drift compensation settings data. Maybe you find some other interesting settings that could be useful when visible in one spot.

tadelv (approved):

PR #65, opened 2026-06-05
Bug audit fixes for dev preview

ODevStudio:
Collects the current bug-audit fixes for a dev preview/tester build.

Scope:
- Firmware hardening for WebSocket, 
BLE/USB protocol bounds, 
WiFi setup/recovery, 
OTA callback handling, 
EEPROM persistence, and quick-boot status reporting.
- Web app fixes for command bindings, 
WebSocket parsing guards, 
CSP, preset parsing, 
and invalid weight handling.

### DIFF
--- a/include/usbcomm.h
+++ b/include/usbcomm.h
@@ -3,6 +3,7 @@
 
 #include "ADS1232_ADC.h"
 #include <math.h>
+#include <string.h>
 
 // Forward declaration of scale object (defined in declare.h)
 extern ADS1232_ADC scale;
@@ -36,6 +37,13 @@ public:
   void (*setTrackingUpdateInterval)(float);
   void (*buttonSquare_Pressed)();
   void (*buttonCircle_Pressed)();
+
+  static const size_t USB_RX_BUFFER_SIZE = 160;
+  static const unsigned long USB_RX_FRAME_TIMEOUT_MS = 30;
+
+  uint8_t usbRxBuffer[USB_RX_BUFFER_SIZE];
+  size_t usbRxLen = 0;
+  unsigned long usbLastByteAt = 0;
 
   uint8_t calculateChecksum(uint8_t *data, size_t len) {
     uint8_t xorSum = 0;
@@ -84,11 +92,24 @@ public:
     return data[frameLen - 1] == calculatePayloadChecksum(data, frameLen - 1);
   }
 
-  size_t boundedFrameLength(size_t len, size_t frameLen) {
-    return len < frameLen ? len : frameLen;
+  size_t fixedFrameLength(size_t len, size_t frameLen) {
+    return len >= frameLen ? frameLen : 0;
   }
 
-  size_t usbCommandFrameLength(uint8_t *data, size_t len) {
+  size_t checksummedOrShortFrameLength(uint8_t *data, size_t len, size_t shortLen, size_t frameLen, bool allowShort) {
+    if (hasChecksummedFrame(data, len, frameLen)) {
+      return frameLen;
+    }
+    if (len < shortLen) {
+      return 0;
+    }
+    if (allowShort || len >= frameLen) {
+      return shortLen;
+    }
+    return 0;
+  }
+
+  size_t usbCommandFrameLength(uint8_t *data, size_t len, bool allowShort) {
     if (data == nullptr || len == 0) {
       return 0;
     }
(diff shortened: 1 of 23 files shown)


platformio.ini
```diff
@@ -22,6 +22,7 @@ framework = arduino
 #board_build.partitions = <min_spiffs.csv>
 board_build.filesystem = littlefs
 monitor_speed = 115200
+upload_port = COM3
```
tadelv: do not specify upload ports this way, some users might have steam hardware on that port, or even have linux port designation

platformio.ini
```diff
@@ -31,7 +32,7 @@ build_flags =
   ; half-open client (connection churn) can't hoard heap. Bounds aggregate heap
   ; growth; complements the WS_BROADCAST_HEAP_FLOOR gate in include/websocket.h.
   -D WS_MAX_QUEUED_MESSAGES=8
-  !python3 git_rev_macro.py
+  !python git_rev_macro.py
```
tadelv: absolutely sure about this one? keep in mind most important is the github action can run

web_apps/Weigh_Save/modules/scale.js
```diff
@@ -161,13 +162,13 @@ async connect(type) {
 
     async _enableNotification() {
         await this.readCharacteristic.startNotifications();
-        this.readCharacteristic.addEventListener('characteristicvaluechanged', this.notification_handler.bind(this));
+        this.readCharacteristic.addEventListener('characteristicvaluechanged', this.notificationHandler);
```
tadelv, line 165: are the characteristics used in the hosted version of Weigh_Save? 
I believe websocket should be used to get weight data

include/websocket.h
```diff
@@ -151,6 +196,14 @@ void processWsPendingCmds() {
   }
   if (mask & WSP_TIMER_STOP)  { stopWatch.stop(); }
   if (mask & WSP_TIMER_ZERO)  { stopWatch.reset(); }
+  if (mask & WSP_SET_SAMPLES) {
+    scale.setSamplesInUse(samplesInUse);
```
tadelv, line 200: which task is this call coming from, the main or the websocket task? possible race?
just want to be sure.

include/webserver.h
```diff
@@ -35,18 +48,28 @@ void startWebServer() {
...
+          if (ssid.length() > WIFI_SETUP_MAX_SSID_BYTES ||
+              pass.length() > WIFI_SETUP_MAX_PASS_BYTES) {
             request->send(400);
             return;
           }
-          String ssid = jsonObj["ssid"];
-          String pass = jsonObj["pass"];
 
-          saveCredentials(ssid, pass);
+          saveCredentialsForRestart(ssid, pass);
```
tadelv: why not forward error from nvs save to api response?

include/usbcomm.h
```diff
@@ -16,6 +17,17 @@ void handleAdsReset(uint8_t mode);
 void sendUsbGyro();
 #endif
 
+static inline uint16_t encodeScaledUInt16(float value, float scale) {
```
tadelv, line 21: we have encode in multiple places if i see this correctly. can we unify the command parsing as well as data serialization for ble and usb?

include/usbcomm.h
```diff
@@ -57,6 +69,124 @@ class MyUsbCallbacks {
     return false;
   }
 
+  uint8_t calculatePayloadChecksum(uint8_t *data, size_t len) {
```
tadelv, line 80: same for checksum, my knowledge about the source might be poor, but we could simplify instead of have duplicates (if we actually have duplicates)

tadelv (commented):
some comments inline. I get the need to sanitize and guard for edge cases, but the work could be split in (at least) half if the packing/unpacking would be unified, it seems.

include/websocket.h
ODevStudio: scale.setSamplesInUse() runs from the main loop via the pending queue, not directly from the WebSocket task

web_apps/Weigh_Save/modules/scale.js
ODevStudio: Yes: when Weigh_Save is served from the scale over HTTP, live weight comes from /snapshot WebSocket. The BLE characteristics are only for direct browser-to-scale connection support.

const ws = new ReconnectingWebSocket(`ws://${window.location.host}/snapshot`);
...
if (jsondata.grams !== undefined) {
  const weight = Number(jsondata.grams);
  ...
  ui.updateWeightDisplay(currentWeight);
  scale.processWeight(currentWeight);
}

include/usbcomm.h
ODevStudio: Yes, there is some duplication in the encode helpers. That would need BLE/USB protocol-helper cleanup as a separate follow-up, with testing.

include/usbcomm.h
ODevStudio: Also for the checksum there are duplications. i keep that in mind

platformio.ini
ODevStudio: Reverted the platformio.ini

platformio.ini
ODevStudio: Reverted the platformio.ini

include/usbcomm.h
tadelv: separate cleanup sounds good

PR #71, opened 2026-06-22
fix: wait for ADC to settle before capturing calibration load

skialpine:
## Problem

Since v3.0.10, calibration fails with **`unstable_load`** on hardware that weighs perfectly. The hardened calibration routine (#ffa4cbb) captures the load reading *before the new ADC lib's smoothed value has settled*.

Root cause is an interaction between two separate changes:
- The 3.0.10 calibration rewrite added a stability gate (reject if read-to-read spread > `max(50, factor×0.5)` counts).
- The #60 `decentespresso/ADS1232_ADC` swap: `setSamplesInUse(16)` clears the lib's ring buffer, so the smoothed value **ramps up** over the next ~16 samples (and again after a load step).

`calibrationCaptureRaw()` returned the first two reads where `validSamples >= 16` — i.e. mid-ramp — so the captured load delta was a fraction of the settled value and the two reads differed wildly.

## Evidence (live capture)

On a unit with a known-good factor (~1133; reads a 50 g weight as 49.95 g, dead stable):

```
reason=unstable_load  zero_raw=8655800 (spread 0)  load_raw=8657823  delta=2023  load_spread=4042
```

- Settled delta for 50 g should be ~56,600 counts; we captured **2,023**.
- Decoding the two load samples: first ≈ 8,655,802 (**still at the zero level**), second ≈ 8,659,844 — a value monotonically ramping up, not scatter around the true mean.
- The zero capture and steady-state weighing were both stable with USB connected, ruling out load-cell noise / USB interference.

## Fix

`calibrationCaptureRaw()` now polls fresh samples until the read-to-read change stays under the stability limit for **3 consecutive reads** (8 s budget) before measuring, so both the zero and load captures wait out the ramp. If the reading never plateaus, it still fails with the unstable reason.

## Verification

Flashed to the affected unit; calibration now succeeds:

```
Calibration accepted: reason=ok  prev=1133.16  candidate=1132.02  verified_g=50.0018
   zero_raw=8655713  load_raw=8712314  delta=56601  zero_spread=3  load_spread=1
```

`delta=56601` (settled), `verified_g=50.00`, new factor matches the prior good factor to 0.1%. One-file change, USB/charging gate unchanged.


ODevStudio:
Included in release PR #73 and shipped in v3.1.11, so closing this PR to avoid a duplicate merge.

Issue #59, opened 2026-05-26
BLE: nimble_host LoadProhibited crash in BLEServer::removePeerDevice (race on m_connectedServersMap)


skialpine:
## Summary

During a sustained multi-protocol soak (WS + HTTP/WS churn + mDNS + USB binary + a connected BT app) on `fix/ws-oom-broadcast` (which carries the OOM fix from PR #58 plus the telemetry from PR #57), the **NimBLE host task crashed** with a `LoadProhibited` exception inside `BLEServer::removePeerDevice` while handling a BLE disconnect event. The scale auto-rebooted (`reset_reason=panic`) and the user's BT app reconnected normally — visible to the operator only via the telemetry `reset_reason` field added in PR #57.

This is a **separate bug class from the OOM crash PR #58 fixes** (which was `bad_alloc` in the WS broadcast path). This one is a long-known **data race in the upstream Arduino-ESP32 BLE library**, not in this firmware's code.

## Evidence (decoded coredump from `/coredump` flash partition)

```
Crashed task: 'nimble_host'              ← BLE stack thread
exccause:     0x1c (LoadProhibitedCause)
excvaddr:     0x60533                    ← garbage pointer (not a valid heap addr)
pc:           0x420202d9  std::_Rb_tree<uint16_t, conn_status_t>::erase + 109
```

Stack (innermost → outermost):

```
#0 std::_Rb_tree<...>::_M_upper_bound     (walking corrupted tree)
#1 std::_Rb_tree<...>::equal_range
#2 std::_Rb_tree<...>::erase
#3 std::map<uint16_t, conn_status_t>::erase(1)
#4 BLEServer::removePeerDevice(conn_id=1, _client=false)
   at framework-arduinoespressif32/libraries/BLE/src/BLEServer.cpp:345
#5 BLEServer::handleGATTServerEvent          (BLEServer.cpp:820)
#6 ble_gap_call_event_cb
#7 ble_gap_conn_broken
#8 ble_gap_rx_disconn_complete               ← BLE disconnect event
#9 ble_hs_hci_evt_disconn_complete
#10..#15  NimBLE host task → vPortTaskWrapper
```

## Root cause

`BLEServer::m_connectedServersMap` (`std::map<uint16_t, conn_status_t>`) is accessed without synchronization from both:
- the NimBLE host task (`addPeerDevice` on connect, `removePeerDevice` on disconnect), and
[...]

PR #72, opened 2026-06-22
chore: pin pioarduino platform to 3.3.9 (BLE fix) + warn when behind stable

skialpine:
## Problem

`platformio.ini` pinned the platform to the **rolling `stable` tag**, whose release asset mutates in place. It silently advanced to arduino-esp32 **3.3.9** on 2026-06-04 — but neither local nor CI builds picked it up:

- The **v3.0.10 release build** resolved `framework-arduinoespressif32 @ 3.3.8` (verified in the [Release CI log](https://github.com/decentespresso/openscale/actions/runs/27934988943) — `PLATFORM: Espressif 32 (55.3.38)`).
- Root cause: `actions/cache@v3` used a **static key** (`${{ runner.os }}-pio`) that restored a months-old cached `stable` platform zip from `~/.platformio/.cache`. PlatformIO served the stale zip (which references the 3.3.8 core) instead of re-downloading the refreshed `stable` asset.

Net effect: the framework version was **neither knowable from `platformio.ini` nor reproducible**, and every shipped release (incl. v3.0.10) carried the 3.3.8 BLE library — which lacks the upstream map-lock fix for the `nimble_host` `LoadProhibited` crash tracked in #59.

## Changes

1. **Pin the platform to the explicit `55.03.39` URL** (= arduino-esp32 3.3.9). This both lands the BLE fix and makes the framework version reproducible. Verified the resolved framework now carries the fix (`BLEServer` guards `m_connectedServersMap` with `m_semaphoreMapAccess`).
2. **Fix the CI cache key** in `release.yml` + `nightly.yml`: key on `hashFiles('platformio.ini')` so a platform/lib pin change busts the stale download cache. The static key is exactly what froze CI on the old zip.
3. **Add `check_platform_freshness.py`** as a pre-build hook (`extra_scripts = pre:`): since we now pin (and must remember to bump), it warns — in the build log and as a GitHub Actions annotation — when the pinned platform is behind the latest pioarduino stable release. **Advisory only**: never fails the build, stays silent offline / rate-limited, and is skippable via `HDS_SKIP_PLATFORM_CHECK=1`.

This is the "pin for reproducibility, but don't let the pin silently rot" approach: deliberate adoption of framework bumps, with a nudge when one is available.

## Verification

```
PLATFORM: Espressif 32 (55.3.39) > Espressif ESP32-S3-DevKitC-1-N8
 - framework-arduinoespressif32 @ 3.3.9
[SUCCESS]
```
- `grep -c m_semaphoreMapAccess …/BLE/src/BLEServer.cpp` → `10` (fix present; was `0` on 3.3.8).
- Freshness hook ran clean with no warning (pinned == latest stable).

## Notes

- The 3.3.9 toolchain (newer GCC) surfaces pre-existing `-Wvolatile` deprecation warnings in `src/wifi_setup.cpp` (volatile `++`). Harmless, not introduced here — flagging in case we want a follow-up to quiet them.
- This addresses the framework-adoption half of #59. Whether to **close** #59 is a judgment call: the fix is now in the build, but the crash is hard to reproduce on demand, so you may prefer to verify under soak first — hence `Refs #59` rather than `Fixes`.

Refs #59

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -22,7 +22,10 @@ jobs:
           path: |
             ~/.cache/pip
             ~/.platformio/.cache
-          key: ${{ runner.os }}-pio
+          # Key on platformio.ini so a platform/lib pin change busts the cache.
+          # A static key froze CI on a stale cached `stable` platform zip (3.3.8)
+          # even after the pin should have resolved a newer framework.
+          key: ${{ runner.os }}-pio-${{ hashFiles('platformio.ini') }}
       - uses: actions/setup-node@v4.1.0
 
       - uses: actions/setup-python@v4
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -63,7 +63,10 @@ jobs:
           path: |
             ~/.cache/pip
             ~/.platformio/.cache
-          key: ${{ runner.os }}-pio
+          # Key on platformio.ini so a platform/lib pin change busts the cache.
+          # A static key froze CI on a stale cached `stable` platform zip (3.3.8)
+          # even after the pin should have resolved a newer framework.
+          key: ${{ runner.os }}-pio-${{ hashFiles('platformio.ini') }}
 
       - uses: actions/setup-python@v4
         with:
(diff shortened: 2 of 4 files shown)


ODevStudio:
Included in release PR #73 and shipped in v3.1.11, so closing this PR to avoid a duplicate merge.

PR #81, opened 2026-07-05
Refactor web app cleanup

ODevStudio:
## Summary

- `e0e1f82` Removes unused web app modules.
- `58919c8` Shares the export download helper.
- `dab81f3` Deduplicates QC weight handling.
- `d5c823d` Keeps the on-device Weigh Save and Dosing Assistant apps on `/snapshot` only by removing hidden BLE/WebUSB fallback code.
- `67fe868` Shares the generated web app stylesheet across Weigh Save, Dosing Assistant, and QC.
- `ceea1bc` Replaces the vendored reconnect script with a small local wrapper.

- `75f40e5` Drops the unused HX711 dependency and guarded HX711 setup code. No current board defines `HX711ADC`.


tadelv (approved):
Approve. Clean refactor — firmware builds green, HX711 removal is fully dead (`HX711ADC` never defined on any board, `DATA_SET` preserved at `config.h:14`, `decentespresso/ADS1232_ADC` pinned-sha dep kept), unity build intact, `processWeight`/`getDoseStatus` rewrites behavior-equivalent, all CSS classes present, no dangling imports of deleted modules.

**MAJOR (fix before/after merge):**
- `web_apps/Weigh_Save/README.md` still describes the removed Web BLE + Web Serial API and references deleted files (`ble_jvs_weigh_save.js`, `serial_api_weigh_save.js`). QC and dosing READMEs were updated; Weigh_Save was missed. Rewrite to `/snapshot`-only so it stops misleading users.

**Minor (non-blocking):**
- Dead `connectButton` / `updateConnectButton` in both `ui-controller.js` (Weigh_Save + dosing) — orphaned when the connect button was removed.
- Stale BLE UUIDs (`CHAR_READ`, `CHAR_WRITE`, `READ_CHARACTERISTIC`, `WRITE_CHARACTERISTIC`) in `dosing_assistant/modules/constants.js` — unused after the BLE removal.
- Duplicate blocks in `shared/app.css` from the three-way CSS merge (no visual regression; dedupe when convenient).

**Suggestion (separate from this PR):** the contract tests (`test_charging_wake_contract.py`, `test_grinder_tcp_contract.py`, `test_soft_sleep_ads_wake.py`) are static source-string assertions — they grep the shape of the code, not its behavior. A refactor that drops the guard clause they're meant to protect still passes. Worth starting to pull testable logic out of the `.ino`/headers into pure, host-compilable units and write real unit tests against those (native toolchain, or a small fake-hardware shim), so the tests exercise the code instead of matching strings. Extracting logic to pure functions would also help with the unity-build header sprawl.

PR #82, opened 2026-07-05
Add HDS Tasmota grinder TCP mode

ODevStudio:
# Summary

Adds HDS support for controlling a grinder through a compatible Tasmota smart plug.

- Select the plug by its MAC address from the HDS OLED menu.
- Set the target, safety margin, and zero range on the scale.
- Tare the empty dosing cup once after each new plug connection.
- The plug turns on when the empty cup is stable and turns off at the calculated cutoff.
- After cup removal and return, the next dose rearms automatically.

The plug-side firmware is maintained in [decentespresso/tasmota-auto-doser](https://github.com/decentespresso/tasmota-auto-doser). See the [complete setup and user guide](https://github.com/decentespresso/openscale/blob/odev/hds-tasmota-grinder/docs/hds-tasmota-grinder.md) for installation, first-use testing, display states, daily operation, and troubleshooting.

# Safety Behavior

- Grinder mode is off by default.
- A newly validated plug connection remains off until a user-requested tare completes and zero is stable.
- Cutoff is `target - safety`, with a minimum cutoff of 5 g.
- OFF is blocked for 1500 ms after weight leaves the positive zero range.
- Weight reaching cutoff above an average 6 g/s is treated as cup/setup mass and remains blocked until tare or return to zero.
- Connection loss while the relay is active enters fail-safe error handling; the plug firmware also turns its relay off when the active TCP session drops.
- Fast OFF uses the plug's emergency `!` command after MAC validation, with normal `OFF` as fallback.

# Discovery And Networking

- Manual `Select Plug` uses `_grinderplug._tcp` mDNS discovery with a bounded TCP fallback.
- Normal weighing does not run full discovery; reconnect uses the cached IP and hostname with short timeouts.
- The selected plug MAC is identity validation, not authentication. This feature is intended for a trusted local WLAN.
- Grinder mode temporarily enables WiFi-on-boot and restores the previous setting when grinder mode is disabled.
- The implementation preserves the default ESP32 WiFi power-save behavior for WiFi/BLE coexistence.

# Testing

- Grinder protocol and firmware integration contracts: `python tools/test_grinder_tcp_contract.py`
- Production parser, cutoff, settings, and adaptive helpers have a host-compilable C++ contract test.
- Existing repository Python tests pass.
- `pio run -e esp32s3` passes.
- Firmware build and upload were verified on an HDS scale over COM5.

# Scope

This PR contains scale-side firmware and documentation only. It does not include plug firmware changes.


tadelv (commented):
Looks solid as a DRAFT and the feature is off by default, so no blocker. Verified clean: no grinder code runs off the loop/core-0 (no AsyncTCP/BLE/USB task touch, no FreeRTOS task spawned), TCP input parsing is bounded (no overflow path), MAC-as-identity/validation scope is correctly documented as trusted-WLAN-only, tare-aware + zero-exit guard + grind-confirmation correctly handle the ADS1232 ramp blip-to-zero (the PR #71 calibration lesson), and the deep-sleep teardown chain flushes grinder settings before `esp32_sleep()`.

**MAJOR to resolve before this leaves DRAFT:**
- `WiFi.setSleep(false)` + `esp_wifi_set_ps(WIFI_PS_NONE)` for the entire grinding session (`grinder_runtime_low_latency.h:49-50`, also `grinder_discovery.h:79-80`). This is the CLAUDE.md #2 WiFi/BLE coexistence footgun — it directly contradicts the explicit comment in `wifi_setup.cpp::connectToWifi` ("Intentionally not calling WiFi.setSleep(false)"). If low-latency grinder TCP genuinely requires `WIFI_PS_NONE`, that needs a maintainer decision and a CLAUDE.md carve-out, and concurrent grinder + BLE use should be gated (right now nothing stops both being active).

**Minor:**
- New globals outside `include/parameter.h`: `b_buttonChordSuppressUntilRelease` (`hds.ino:27`), `b_grinderMenuDirectEntry` (`menu.h:217`), `grinderSettings`/`grinderRuntime` (`grinder_runtime.h:112-113`). No cross-task access so no thread risk, but CLAUDE.md says globals belong in `parameter.h` for discoverability.
- Forcing WiFi-on-boot is a one-way ratchet: `grinderOn` writes `b_wifiOnBoot=true` to EEPROM but `grinderOff` doesn't restore prior state. Once grinder is enabled, WiFi-on-boot stays on even after grinder is disabled. Probably intentional, but it's an undocumented user-facing side effect — document or restore.
- 12s blocking discovery poll (`delay(250)` loop in `grinderEnsureWifiReadyForDiscovery`). Acceptable since it's menu-initiated, just noting it blocks `processWsPendingCmds` / WS drain during that window.

**Suggestion (separate from this PR):** `test_grinder_tcp_contract.py` is static source-string matching. As the grinder protocol/state-machine grows, pulling it into pure host-compilable units with real unit tests would catch behavior regressions the grep tests can't — especially for the cutoff/arming/zero-exit logic where the interesting bugs live.

ODevStudio:
Thanks for the detailed review. The major and minor findings have been addressed:

- Removed all grinder-side WiFi power-save overrides. The firmware no longer calls `WiFi.setSleep(false)`, `esp_wifi_set_ps()`, or `WIFI_PS_NONE`; normal ESP32 WiFi/BLE coexistence remains active. Contract tests enforce their absence from grinder runtime and discovery code.
- Moved the button/menu globals into `include/parameter.h`. `grinderSettings` and `grinderRuntime` are declared there with `extern` and defined alongside their complete types in `grinder_runtime.h`.
- Grinder enable now stores the previous WiFi-on-boot setting, and `Grinder Off` restores it instead of leaving WiFi permanently enabled.
- Manual discovery remains intentionally blocking only inside `Select Plug`. Normal weighing never performs full mDNS/subnet discovery and uses bounded cached-IP/hostname attempts instead.
- Added a host-compilable C++ contract test using the production parser, cutoff, settings, adaptive-safety, guard-reset, setup-mass, and malformed-response helpers. The remaining Python source checks cover firmware integration and call ordering.

Additional safety work now requires a user tare after each new validated plug connection, keeps the 1500 ms zero-exit guard, rejects implausible setup-mass rises above 6 g/s, keeps cutoff at least 5 g, and documents setup, display states, daily use, and troubleshooting in `docs/hds-tasmota-grinder.md`.

tadelv (commented):
## Follow-up review (adding to tadelv's earlier review)

The earlier review points (WiFi.setSleep, globals, WiFi-on-boot ratchet, test quality) are addressed in the current head. Adding cross-PR observations:

### Blocking: EEPROM conflict with #88 (NVS migration)
`grinderOn()` and `grinderOff()` still use the old EEPROM API:
```cpp
EEPROM.put(i_addr_enableWifiOnBoot, true);
EEPROM.commit();
```
PR #88 removes `i_addr_enableWifiOnBoot` from `parameter.h` and replaces all EEPROM access with `storagePutBool(KEY_WIFI_BOOT, ...)`. These will conflict on merge. After #88 lands, rebase these calls to:
```cpp
storagePutBool(KEY_WIFI_BOOT, true);
```

The grinder's own NVS namespace (`grinder`) is correctly isolated and won't conflict — only the WiFi-on-boot EEPROM writes need updating.

### Note: File sizes
`grinder_runtime.h` (909 lines) and `menu.h` (now ~1776 lines with grinder additions) are large. Not a blocker, but if #89's proposed "800-line ceiling" guideline is adopted, these would violate it. Consider splitting `grinder_runtime.h` — the state machine + TCP I/O is cleanly separable from discovery + menu helpers.

### Good: Dual-tick design
The `grinderRuntimeFreshWeightTick` (tracking-compensated weight, called from `pureScale()` for low-latency cutoff) + `grinderRuntimeTick` (stable output, called from `loop()` for state transitions) with sequence-number dedup is a well-considered design. Gives sub-loop cutoff latency without destabilizing the state machine.

### Good: `grinderRuntimeLocksScaleSampling()`
Correctly prevents `setSamplesInUse()` changes during grinding, which would trigger the ADC ring-buffer ramp documented in CLAUDE.md (PR #71 lesson). Also blocks calibration sample window changes.

ODevStudio:
The order and how is already mentioned in [PR](https://github.com/decentespresso/openscale/pull/88)

ODevStudio:
Rebased onto main after #88. Grinder On/Off now persist WiFi-on-boot through the HDS NVS storage API instead of runtime EEPROM; grinder settings remain in the separate grinder Preferences namespace. Storage, grinder, calibration, soft-sleep, charging-wake, native MSVC contract, ESP32-S3 build, and physical grinder behavior were validated.

PR #86, opened 2026-07-07
feat(power): configurable deep-sleep GPIO strategy with build flags

Sofronio:
Add two compile-time flags (both enabled by default) to control GPIO
state during deep sleep, reducing sleep current from 75 μA to 60 μA.

**-D SLEEP_GPIO_INPUT**
PWR_CTRL-domain pins set to INPUT (float) during deep sleep
instead of OUTPUT LOW + gpio_hold.  Pins return to their default
reset state naturally; no gpio_hold needed.

**-D SLEEP_ISOLATE_UNUSED**
rtc_gpio_isolate on 11 unused pins (JTAG 39-43, PSRAM 35-37,
unconnected 22, 23, 38).  Disconnects the pad driver internally
so no current can flow regardless of external pull-ups.

**Measured deep-sleep current (V8.1, running ~78 mA):**

| Mode | Sleep current |
|---|---|
| OUTPUT LOW + no isolate (original) | 75 μA |
| INPUT float + isolate | 60 μA |

Wake-up path (setup) always calls gpio_hold_dis + gpio_deep_sleep_hold_dis
regardless of flags — harmless no-op when holds were not set.  Unused pins
remain isolated after wake-up (not de-initialized since firmware never
references them).

### DIFF
--- a/include/config.h
+++ b/include/config.h
@@ -119,6 +119,12 @@
 #define SCALE_A0 -1
 #define HX711_SCL 12
 #define HX711_SDA 11
+	// Unused pins isolated during deep sleep (INPUT + rtc_gpio_isolate):
+	//   22, 23, 38          — unconnected pads
+	//   35, 36, 37          — octal PSRAM (powered off during sleep)
+	//   39, 40, 41, 42, 43  — JTAG header
+	// See include/power.h esp32_sleep() for the isolation array.
+	#define SLEEP_ISOLATE_GPIO_CNT 11
 #endif
 
 
--- a/include/power.h
+++ b/include/power.h
@@ -219,10 +219,46 @@ void esp32_sleep() {
   // A partially-powered ADS1232 / REF5025 / ADS1115 may not trigger POR
   // on the next wake-up, causing latch-up and the "stuck at 0g" bug.
   //
-  // Fix: drive every GPIO in the PWR_CTRL domain LOW (or INPUT for DOUT
-  // pins) and enable gpio_hold so the state persists through deep sleep.
-  // I2C is driven LOW to sink the external pull-up current safely instead
-  // of letting it flow through unpowered-device ESD diodes.
+  // Default (SLEEP_GPIO_INPUT not defined): drive every GPIO in the
+  // PWR_CTRL domain LOW (or INPUT for DOUT pins) and enable gpio_hold so
+  // the state persists through deep sleep.  I2C is driven LOW to sink the
+  // external pull-up current safely instead of letting it flow through
+  // unpowered-device ESD diodes.
+  //
+  // When SLEEP_GPIO_INPUT is defined: configure all PWR_CTRL domain GPIOs
+  // as INPUT with no pull-up/pull-down.  This saves the pull-up sink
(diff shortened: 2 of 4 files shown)


tadelv (changes requested):
Request changes. Two default-on flags whose primary mechanisms don't hold up on this hardware.

**BLOCKER:**
- `SLEEP_ISOLATE_UNUSED` is a no-op. ESP32-S3 exposes only GPIO 0-21 as RTC GPIOs; pins 22, 23, 35-43 are digital-only. `rtc_gpio_isolate()` returns `ESP_ERR_INVALID_ARG` and does nothing on them. The isolation list `{22, 23, 35, 36, 37, 38, 39, 40, 41, 42, 43}` is entirely non-RTC, so the feature ships inert — only the `pinMode(INPUT)` calls have any effect, and those pins are unused/high-Z already. Either target genuinely-unused RTC GPIOs (most of 0-21 are in use on V8_1, though) or drop the flag.

**MAJOR:**
- The isolation pin list is hardcoded for V8_1 but the flag is global in `platformio.ini`. V7_4 uses pin 38 for `BUZZER` and pin 35 for `SCALE2_PDWN`; V7_5 uses pin 39 for `SCALE2_PDWN`. `power.h` comments say isolated pins stay isolated after wake — so building any non-V8_1 board with the flag on (which `platformio.ini` does unconditionally) breaks the buzzer / SCALE2 power-down after the first sleep. Make the list board-derived from `config.h`, or scope the flag per board.
- `SLEEP_GPIO_INPUT` defaults ON for all boards and reintroduces the documented stuck-at-0g / ESD-backfeed path. The existing I2C LOW-drive (and CLAUDE.md) exists specifically to sink pull-up current away from the unpowered ADS/gyro ESD diodes. Floating those pins lets the always-on pull-ups back-feed device VDD — the latch-up the prior fix prevented. The PR comment even admits this tradeoff. Should be opt-in per board (default off), not a global default.
- The 15 µA saving contradicts the claimed mechanism: floating I2C would *increase* sleep current (pull-ups into VDD), not decrease it. Needs on-device validation that the 60 µA figure actually holds and that it doesn't reproduce the stuck-at-0g bug across sleep/wake cycles.

**MINOR:**
- `platformio.ini` comment is self-contradictory: "Uncomment the line below to switch…" but both `-D SLEEP_ISOLATE_UNUSED` and `-D SLEEP_GPIO_INPUT` are already uncommented. Describes a "default" the next lines override.
- `SLEEP_ISOLATE_GPIO_CNT` is dead code — defined in the V8_1 block of `config.h` but never referenced (the array length is hardcoded in `power.h`). Remove or actually use it for the loop bound.

**Good parts (keep):** threading is clean (`esp32_sleep` loop-only, no AsyncTCP touch), `gpio_hold_dis` + `gpio_deep_sleep_hold_dis` in the wake path is a harmless no-op when holds weren't set, and reordering `gpio_hold_dis` before the `pinMode`/`digitalWrite` restore in `setup()` is actually more correct than before (the old code wrote HIGH while hold was still latched LOW).

Sofronio:
I tested this PR, focusing on the 3v3P pin, it was 0.35v, so it'll break the 0g fix. I'll close it now.

PR #88, opened 2026-07-10
feat: migrate HDS settings to NVS Preferences

ODevStudio:
## Summary
- move HDS settings from EEPROM offsets to typed Preferences keys in the HDS-only `hds` namespace
- migrate legacy EEPROM values once without erasing the rollback-compatible blob
- keep WiFi and OTA namespaces unchanged; 

## Validation
- `python tools/test_storage_migration_contract.py`
- `python tools/test_calibration_validation.py`
- `python tools/test_soft_sleep_ads_wake.py`
- `python -m unittest test_ads_debug_protocol.py test_decode_ads_debug.py` from `tools/`
- `pio run -e esp32s3`
- live COM5 migration, reboot persistence, reversible NVS write, firmware rollback, and migration-firmware restore

## Open PR compatibility
- PRs #82 and #87 are not included because they are still open; after #88 merges, both should be rebased onto `main`.
- PR #82 must replace its remaining HDS WiFi EEPROM writes with `storagePutBool(KEY_WIFI_BOOT, ...)`; grinder settings remain in the separate `grinder` namespace.
- PR #87 keeps OTA state in the separate `ota_fs` and `ota_verify` namespaces and requires no HDS schema changes.

## Notes
- blank-device NVS erase was not performed on the connected scale

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -116,6 +116,10 @@ if (info->final && info->index == 0 && info->len == len && info->opcode == WS_TE
 }
 ```
 
+## Persistent settings
+
+Scale settings use the HDS-specific `hds` NVS namespace and SensorBasket-style named keys from `include/storage.h`. WiFi and OTA keep separate namespaces. The legacy EEPROM blob is read only while schema migration is incomplete and is retained for firmware rollback.
+
 ## Style conventions
 
 | Prefix | Meaning |
--- a/include/ble.h
+++ b/include/ble.h
@@ -71,7 +71,7 @@ class MyServerCallbacks : public BLEServerCallbacks {
     deviceConnected = false;
     bleState = DISCONNECTED;
 #ifdef BUZZER
-    EEPROM.get(i_addr_beep, b_beep);  //read buzzer state after disconnect
+    b_beep = storageGetInt(KEY_BEEP, 1);
 #endif
     b_u8g2Sleep = false;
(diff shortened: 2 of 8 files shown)


tadelv (commented):
## Review: NVS Migration

Clean migration — verified all legacy EEPROM address offsets in `storage.h` match the original `parameter.h` calculations. Good contract test. This should merge first since #87 and #82 both need to rebase onto it.

### Suggestion (non-blocking)
`storagePutString()` round-trip verification uses `"\x01"` as a sentinel:
```cpp
return settingsPreferences.isKey(key) &&
       settingsPreferences.getString(key, "\x01") == value;
```
If a legitimate stored value were ever `"\x01"`, this would falsely report failure. In practice the welcome text is always longer, so this won't fire — but consider trusting `putString()` directly (it returns bytes written) or using a less collision-prone sentinel.

### Note for downstream PRs
After this merges:
- **#82** must replace `EEPROM.put(i_addr_enableWifiOnBoot, ...)` in `grinderOn()`/`grinderOff()` with `storagePutBool(KEY_WIFI_BOOT, ...)`
- **#87** must replace `EEPROM.begin(512)` / `readBoolEEPROMWithValidation()` in `setup()` with `storageInit()` / `storageGetBool()`
- The `test_storage_migration_contract.py` assertion `"EEPROM" not in source` for 5 files is a good guard — extend it to cover any new files added by those PRs.

PR #90, opened 2026-07-11
fix: lock partition table to default_8MB.csv

tadelv:
## Summary

Replace the commented-out placeholder partition table with the upstream arduino-esp32 `default_8MB.csv`.

### Partition layout

| Partition | Type | Size | Purpose |
|-----------|------|------|---------|
| nvs | data/nvs | 20 KB | Non-volatile storage |
| otadata | data/ota | 8 KB | OTA slot tracking |
| app0 | app/ota_0 | 3.2 MB | First OTA app slot |
| app1 | app/ota_1 | 3.2 MB | Second OTA app slot |
| spiffs | data/spiffs | 1.5 MB | LittleFS (web assets) |
| coredump | data/coredump | 64 KB | Crash dumps |

### Rationale

HDS < 2.0 uses **16 MB flash**; HDS 2.0 will use **8 MB flash**. Pinning the partition table to 8 MB now ensures the HDS 2.0 flash-size transition needs **no partition table change** — the firmware already fits in 8 MB. A partition table change on existing deployed devices requires erasing flash + re-flashing (cannot be done via OTA alone).

**Trade-off:** wastes the upper 8 MB on current HDS 1.x hardware (16 MB flash). This is an intentional forward-compatibility decision.

Reference: https://github.com/espressif/arduino-esp32/blob/master/tools/partitions/default_8MB.csv

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -40,6 +40,20 @@ jobs:
           pio run -e ${{ matrix.board }} 
           pio run -e ${{ matrix.board }} -t buildfs
 
+      - name: Assert vendored partition table matches framework
+        run: |
+          VENDORED=partitions/default_8MB.csv
+          UPSTREAM=~/.platformio/packages/framework-arduinoespressif32/tools/partitions/default_8MB.csv
+          if ! diff "$VENDORED" "$UPSTREAM" > /dev/null 2>&1; then
+            echo "::warning::Vendored $VENDORED differs from the framework-owned default_8MB.csv."
+            echo "This may be intentional (we pinned to an older layout) or the framework"
+            echo "has changed its partition table. Review the diff and decide whether to"
+            echo "update the vendored copy:"
+            diff "$VENDORED" "$UPSTREAM" || true
+          else
+            echo "Vendored partition table matches framework default_8MB.csv."
+          fi
+
       - name: "Run slugify"
         id: slugify
         uses: eltimn/slugify-action@v2.0.2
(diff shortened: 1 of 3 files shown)


ODevStudio (changes requested):
Request changes: Please vendor the exact 8 MB partition CSV in this repository instead of referencing the framework-owned default_8MB.csv. The current platform pin makes the result reproducible today, but a later platform upgrade can change that built-in file while preserving its filename. Since stable offsets are the purpose of this PR, the layout itself should be repository-controlled and protected by a CI assertion.

Also, the current board definition already defaults to default_8MB.csv, so this PR is presently a no-op rather than a partition transition. Please clarify that in the description and correct the contradictory “HDS <= 2.0” versus “HDS 2.0” wording.

tadelv (commented):
Thanks for the review — all three points addressed in 029614c:

1. **Vendored the CSV** — copied the exact upstream `default_8MB.csv` into `partitions/default_8MB.csv` and pointed `board_build.partitions` at the vendored path instead of the framework-owned file.

2. **CI assertion** — added a non-blocking step to `nightly.yml` that diffs the vendored CSV against the framework copy and emits a `::warning` annotation when they diverge (with the diff and guidance). Non-blocking by design, since divergence is expected when we deliberately pin to a platform version and upstream moves on.

3. **Clarified the no-op** — updated the comment to acknowledge the board already defaults to this layout, and fixed the "HDS ≤ 2.0" → "HDS 1.x / HDS 2.0" wording.

PR #113, opened 2026-07-28
feat(wifi): make the mDNS name configurable, and stop orphaning DNS-SD instances

skialpine:
> **Where this came from:** I now have two Half Decent Scales on one network. That is the whole origin of this PR — everything here is a problem you only meet with a second scale on the LAN.

## The problem

Every scale asks for the same identity. `src/wifi_setup.cpp` hardcoded all three strings:

```cpp
MDNS.begin("hds");
MDNS.setInstanceName("Half Decent Scale");
WiFi.setHostname("hds.local");
```

The mDNS stack does resolve the collision — the loser gets suffixed to `hds-2.local` — so both scales stay reachable. But:

- **Which scale is `hds` vs `hds-2` follows boot order.** Power them up the other way round and they swap addresses, so a bookmark or app config cannot pin a specific scale.
- **`hds-2` tells you nothing.** Not which scale is by the grinder.
- **Hostname and instance name are probed independently.** Observed live with two scales up: `192.168.10.145` was at `hds-2.local` labelled `Half Decent Scale`, while `192.168.10.241` was at `hds.local` labelled `Half Decent Scale-2`. Each won one race and lost the other, so address and browser label disagreed on both units at once.

Also, `WiFi.setHostname("hds.local")` passed a `.local` suffix to a DHCP hostname, where it does not belong.

## What this changes

A device name in the NVS `wifi` namespace under `mdns_name`, defaulting to `hds`. **A scale that is never renamed behaves exactly as it does today** — same `hds.local`, same `Half Decent Scale` instance name, no migration, no schema bump.

The name drives:

| Target | Default | Renamed to `kitchen` |
| --- | --- | --- |
| `MDNS.begin()` | `hds` | `kitchen` |
| `WiFi.setHostname()` | `hds` | `kitchen` |
| DNS-SD instance | `Half Decent Scale` | `Half Decent Scale (kitchen)` |
| TXT record | `name=hds` | `name=kitchen` |

Renaming is `POST /setup/name`, driven by a "Scale Name" field in the WiFi setup area of the on-device web UI. The name also appears in the `/snapshot` status frame as `mdns_name`, and on the OLED WiFi status screen so it can be recovered without network access.

## Second thing: stale DNS-SD instances

`stopWifi()` never called `MDNS.end()`, and that call is what emits the DNS-SD goodbye. So every deliberate reboot orphaned its instance registration: the PTR record sat in every resolver cache on the LAN for its full TTL — 75 minutes by DNS-SD convention, against 2 minutes for SRV/A — which is exactly why a browse would list instances that never resolve.

A `_decentscale._tcp` browse here returned four instances for two scales, two of them ghosts.

`stopWifi()` now withdraws first, while the radio is still up. Every deliberate teardown routes through it — remote and USB reset, rename and wifi-setup reboots, deep sleep — so one call site covers them all. The heap-watchdog `esp_restart()` stays best-effort by nature.

**This bug predates the naming feature** (the `Half Decent Scale-2` ghost came from plain 3.1.12) and is separable if you would rather take it on its own.

Worth being explicit about what it does not fix: a goodbye is one unacknowledged multicast, and a crash, flat battery, or unplug sends nothing at all. Clients still have to treat an unresolvable browse hit as "not a scale".

## Name rules

`include/mdns_name.h`, deliberately Arduino-free so the `native` env can test it: 1-24 characters of `[a-z0-9-]`, no leading or trailing hyphen, trimmed and lowercased, and an empty value restores the default `hds`.

Reposting the name a scale already has is a no-op — no NVS write, no restart, response says `"restarting":false`. The web UI prefills the field, so without that check a stray submit would reboot the scale mid-shot, and a repeated POST of the current name would hold it in a reboot loop.

## Firmware constraints respected

- The `/setup/name` handler runs on the AsyncTCP task, so it touches no mDNS entry point and publishes no `String` the main loop reads. It writes NVS and queues the existing deferred restart, the same shape `/setup/wifi` already uses.
- `WiFi.setHostname()` is only read before association, which is why a rename restarts rather than re-advertising live.
- The OLED name line is drawn inside `showWifiStatus()`'s existing `firstPage()`/`nextPage()` loop — no nested page loop.
- The name lives in the `wifi` namespace, not `hds`, so the settings schema and its migration contract are untouched.

## Verification

Host-side:

```
pio test -e native -f test_mdns_name     8/8 passed
python tools/test_mdns_name_contract.py  passed
python tools/test_ai_docs_contract.py    passed
python tools/test_storage_migration_contract.py  passed
pio run -e esp32s3                       SUCCESS (RAM 18.4%, Flash 49.8%)
```

`nightly.yml` now runs the new contract check and a `native-tests` job — neither the unit tests nor the contract script had any CI coverage before, since no workflow ran `pio test`.

On hardware (two scales on one LAN):

- Default boot answers `hds.local`, TXT `name=hds`, status `mdns_name='hds'`.
- `{"name":"  Kitchen  "}` → `200 {"name":"kitchen","restarting":true}`; after reboot `kitchen.local` resolves, instance is `Half Decent Scale (kitchen)`, TXT `name=kitchen`, status agrees.
- `{"name":""}` restores `hds`.
- Rejected with `400`: `my scale`, `hds.local`, `-bad`, and a body with no `name`.
- Three rapid posts of the current name: uptime ran 80000 → 90000 ms unbroken, heap flat — no reboot.
- Passive 5353 capture across a rename shows the withdrawal working:

```
[16.8s] PTR ttl=0     Half Decent Scale (hdstest)     <<< goodbye, before the radio drops
[25.9s] PTR ttl=4500  Half Decent Scale (hdstest2)
```

  A follow-up browse listed only live instances — no `hdstest` ghost.

One observation for the record: ESP-IDF sends the goodbye for the PTR record only, not SRV/TXT/A. That is the right one to withdraw, since PTR carries the 4500 s TTL that causes the ghosts while SRV/A expire on their own in 120 s.

### DIFF
--- a/include/menu.h
+++ b/include/menu.h
@@ -3,6 +3,7 @@
 
 #include "esp32-hal.h"
 #include "parameter.h"
+#include "mdns_name.h"
 #include "wifi_setup.h"
 #include "grinder_runtime.h"
 #include <string.h>
@@ -332,17 +333,9 @@ void showWifiStatus() {
   String ip = WiFi.isConnected() ? WiFi.localIP().toString() : "0.0.0.0";
   const char *status = WiFi.isConnected() ? "Enabled" : "Disabled";
 
-  // 6x12 leaves ~16 chars after the "Name:" label on one line; a name up to
-  // MDNS_NAME_MAX_CHARS wraps onto a second, label-free line so every
-  // accepted name stays readable here without network access.
-  const char *name = wifiDeviceName();
-  const size_t nameLine1Chars = 16;
-  size_t nameLen = strlen(name);
-  size_t line1Len = nameLen < nameLine1Chars ? nameLen : nameLine1Chars;
-  char nameLine1[nameLine1Chars + 1];
-  memcpy(nameLine1, name, line1Len);
-  nameLine1[line1Len] = 0;
-  const char *nameLine2 = name + line1Len;
+  char nameLine1[MDNS_NAME_OLED_LINE1_BYTES];
+  const char *nameLine2 =
+    mdnsNameSplitOledLine1(wifiDeviceName(), nameLine1, sizeof(nameLine1));
 
   u8g2.firstPage();
   do {
--- a/include/menu.h
+++ b/include/menu.h
@@ -345,6 +345,9 @@ void showWifiStatus() {
     u8g2.drawStr(0, 44, "IP:");
     u8g2.drawStr(40, 44, ip.c_str());
 
+    u8g2.drawStr(0, 60, "Name:");
+    u8g2.drawStr(40, 60, wifiDeviceName());
+
   } while (u8g2.nextPage());
   delay(1000);
   while (b_showWifiData) {
--- a/include/menu.h
+++ b/include/menu.h
@@ -5,6 +5,7 @@
 #include "parameter.h"
 #include "wifi_setup.h"
 #include "grinder_runtime.h"
+#include <string.h>
 const char *weights[] = { "Exit", "50g", "100g", "200g", "500g", "1000g" };
 const float weight_values[] = { 0.0, 50.0, 100.0, 200.0, 500.0, 1000.0 };
 bool b_showAbout = false;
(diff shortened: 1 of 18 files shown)


ODevStudio (changes requested):
The configurable naming work looks solid, and the hardware capture gives good evidence that `MDNS.end()` fixes the stale DNS-SD records. Two issues still need changes before merge:

**1. OTA reboots bypass the withdrawal path**

`stopWifi()` now performs the mDNS withdrawal, but the OTA success paths do not reach it:

- Pull OTA calls `ESP.restart()` after the firmware update and after the LittleFS update.
- ElegantOTA enables automatic reboot, so a successful upload bypasses `reset()` as well.

Please disable ElegantOTA auto-reboot and queue successful OTA resets through the existing deferred main-loop reset mechanism. Pull OTA should use the same path after a successful update. This keeps mDNS and other hardware cleanup on the main task rather than calling it from AsyncTCP or the OTA task.

Please also audit `hdsOtaRollback()`. `esp_ota_mark_app_invalid_rollback_and_reboot()` can run after the pending LittleFS flow has brought WiFi up, so it is another direct reboot that may skip withdrawal. Perform the shared teardown from a safe task before invoking the rollback reboot, or document why mDNS cannot be active there.

**2. The OLED clips valid names**

Names up to 24 characters are accepted, but the WiFi screen draws the value at x=40 with the 6x12 fixed-width font. A 128-pixel display leaves room for only 14 characters, so names such as `kitchen-coffee-scale` lose the distinguishing suffix.

Please make every accepted name recoverable from the display. A smaller full-width line, wrapping, scrolling, or a separate page would work. Reducing the accepted limit is also possible, but it would need matching API, UI, documentation, and test changes.

Please add focused coverage for the OTA reboot routing and manually validate a 24-character name on the supported 128x64 displays. The current shutdown contract only checks `stopWifi()` and therefore passes while the OTA paths still bypass it.

Once these two paths are covered, the PR should meet its stated naming and stale-record goals.

ODevStudio:
Follow-up after checking the pinned Arduino-ESP32 3.3.11 implementation:

**DHCP hostname setup order**

`setupWifi()` currently calls:

```cpp
WiFi.config(INADDR_NONE, INADDR_NONE, INADDR_NONE, INADDR_NONE);
WiFi.setHostname(params.getMdnsName());
```

`WiFi.config()` calls `STA.begin()`. When STA is enabled, the framework copies the current default hostname into the network interface. The later `WiFi.setHostname()` only changes the stored default and does not update the active STA interface.

As a result, mDNS may advertise `kitchen.local` while DHCP still receives the generated hostname, such as `esp32s3-xxxxxx`.

Please reverse the calls:

```cpp
WiFi.setHostname(params.getMdnsName());
WiFi.config(INADDR_NONE, INADDR_NONE, INADDR_NONE, INADDR_NONE);
```

A focused contract check should ensure `setHostname()` remains before the first operation that starts STA.

skialpine:
Thanks for the review. Comments addressed.

include/menu.h
```diff
@@ -331,19 +332,37 @@ void showWifiStatus() {
   String ip = WiFi.isConnected() ? WiFi.localIP().toString() : "0.0.0.0";
   const char *status = WiFi.isConnected() ? "Enabled" : "Disabled";
 
+  // 6x12 leaves ~16 chars after the "Name:" label on one line; a name up to
+  // MDNS_NAME_MAX_CHARS wraps onto a second, label-free line so every
+  // accepted name stays readable here without network access.
+  const char *name = wifiDeviceName();
+  const size_t nameLine1Chars = 16;
```
ODevStudio: include/menu.h:339 — OLED wrapping still drops characters

All supported displays are 128×64. The name is drawn at x=40 using the fixed-width u8g2_font_6x12_tr, leaving 88 pixels—space for only 14 complete six-pixel characters.

The implementation splits after 16 characters, draws those 16 at x=40, then starts the second line at character 17. Consequently, characters 15–16 are clipped from the first line and absent from the second. A valid 24-character name is therefore still not fully recoverable from the OLED.

For example, abcdefghijklmnopqrstuvwx would effectively show characters 1–14 and 17–24, losing o and p.

Required fix: split at 14 characters, or calculate the largest prefix whose measured rendered width fits between x=40 and the display edge. Add a focused check or manually verify a 24-character name.

The earlier OTA-reset routing, rollback teardown, and DHCP-hostname ordering findings are otherwise addressed in the current head commit. The current CI workflow runs also completed successfully.

ODevStudio (changes requested):

skialpine:
Thanks for the second review. Think I resolved the issue this time.

PR #122, opened 2026-08-11
Add optional Pressensor pressure sensor mode

csanadymiklos:
## What

Adds Pressensor as the first approved firmware patch plugin for OpenScale.

The production firmware source tree remains unchanged. The implementation is stored in `plugins/pressensor/patches/main.patch` and is applied only in an isolated custom-build checkout when the user selects the `pressensor` plugin. The package declares no WiFi, WebServer, or runtime LittleFS dependency.

After application, the plugin adds an optional BLE-central mode that connects to a selected Pressensor and displays pressure, weight, flow rate, and shot time. It remains contained behind `HDS_ENABLE_PRESSENSOR` and provides the dedicated validation environment `esp32s3-pressensor`.

## Plugin package

- ID: `pressensor`
- Version: `1.0.0`
- Firmware ref: `main`
- Patch: `plugins/pressensor/patches/main.patch`
- Runtime assets: none
- Required optional features: none
- Conflicts: none

The custom-build workflow detects patch plugins changed by a pull request and compiles only those plugin/ref pairs. For this PR it produces exactly `plugin (pressensor, main)`; unrelated plugins and the old feature matrix are not rebuilt.

## Grind by weight selector

The existing built-in grinder implementation remains unchanged and is not converted into a patch. A manifest-only `grind-by-weight` package exposes it under Approved plugins in the configurator.

- Public name: `Grind by weight`
- Internal compatibility feature: `grinder`
- Patch and runtime assets: none
- Resolved dependencies: WiFi and mDNS
- Existing flags and validation environment remain `HDS_FEATURE_GRINDER`, `HDS_ENABLE_GRINDER`, and `esp32s3-grinder`

The internal feature card is hidden, the plugin selection resolves the existing compile gate, and the OLED menu and user documentation use the public name.

## Behavior

- Sensor selection is explicit; an empty saved MAC does not auto-connect to arbitrary PRS devices.
- Zeroing remains an explicit menu action and is not performed automatically on reconnect.
- BLE callback data crosses to the main loop through synchronized state.
- Subscription failures, stale notification streams, disconnects, and late events recover without letting an old connection attempt overwrite a newer one.
- Auto start is pressure-only; auto stop uses flow and collected weight.
- Flush and dry-run guards prevent an abandoned timer.

## Resource measurements

Measured against `esp32s3` at the same base revision:

- Application flash: +14,776 bytes
- Total image: +15,960 bytes
- Static RAM: +640 bytes
- LittleFS: +0 bytes

These values fit the declared Pressensor package budgets of 32,768 bytes flash, 2,048 bytes static RAM, and 0 bytes LittleFS.

## Validation

- Standard `esp32s3` build: passed
- Pressensor patch application and contract test: passed
- Isolated `esp32s3-pressensor` build: passed locally and in GitHub Actions
- Exact `grind-by-weight` custom selection build: passed locally
- Dynamic changed-plugin detection: passed
- OTA, native, AI-documentation, catalog, custom-build, and Worker contracts: passed
- Configurator verified on desktop and mobile
- OLED menu smoke test on a connected HDS scale: passed
- Earlier Pressensor implementation was tested by the contributor with real hardware

The final artifact produced by the public configurator will be sent to the plugin contributor for a hardware validation pass after merge. Any hardware finding will be handled as a versioned plugin update rather than placing the implementation in the normal firmware tree.


ODevStudio (changes requested):
Thanks for putting this together. I plan to add it later as a plugin just like the grinder functionality — the feature is nicely isolated behind `HDS_ENABLE_PRESSENSOR`, and the overall structure is easy to follow. The dedicated environment, menu integration, and shot-state handling all make sense.

I found a few BLE edge cases that I think are maybe worth addressing but its up to you:

1. **Avoid zeroing the Pressensor automatically on reconnect.**
   The current connection flow calls `pressensorZeroNow()` as soon as notification setup completes. If the BLE link happens to reconnect while the machine is under pressure, that would make the current pressure the new zero point and shift all subsequent readings. I think zeroing should stay behind the explicit “Zero Sensor” action, or only happen at a point where zero pressure is guaranteed.

2. **Synchronize the BLE callback state with the main loop.**
   The GAP/GATT/scan callbacks and `pressensorLinkTick()` both read and modify `pressensorLink`. Some individual fields are `volatile`, but the shared state as a whole is not synchronized. This also makes late callbacks from an old connection attempt potentially affect a newer one. It would be safer to move callback results through a protected snapshot/queue/critical section, ideally with a connection generation or handle check.

3. **Recover if notifications never start or become stale.**
   The code enters `PRESSENSOR_LINK_STREAMING` immediately after starting the CCCD write, before confirming the subscription succeeded. `pressensorStoreBar(0.0f)` also makes the link appear healthy for the first few seconds. If notifications never arrive, or later stop arriving, `pressensorStreaming()` becomes false but the state machine stays in `PRESSENSOR_LINK_STREAMING` indefinitely. A subscription callback plus a stale-stream reconnect path would make this much more robust.

4. **Don’t auto-connect to any PRS device when no sensor is selected.**
   With an empty saved MAC, `pressensorLinkBegin()` enables `matchAnyDevice`. That means enabling Pressensor mode before selecting a sensor can connect to whichever PRS device happens to be nearby, even though the UI says “Select Sensor.” I’d prefer leaving the link idle until a sensor has explicitly been selected.

Apart from those points, the feature looks thoughtfully contained and the PR description does a good job explaining the design choices. 

Edit: did the changes. hope its ok. will let you know when its ready for testing but could get it to run on my side

ODevStudio (approved):
The requested BLE fixes are addressed. The Pressensor patch build, standard firmware build, catalog contracts, and Grind by weight selector checks are green.

PR #146, opened 2026-08-13
fix(ble): deliver notifications to iOS 9 clients (Decent Scale on old iPads)

johnbuckman:
## Problem

On a jailbroken **iPad mini 1 (iOS 9.3.5)**, de1app connects the HDS/Decent Scale but never receives weight — the de1app watchdog reports **"abandoning scale updates"**. The Atomax Skale works on the same iOS 9 device, and HDS works on macOS / modern iOS. So it is HDS-specific *and* iOS-9-specific.

## How it was found

Diagnosed both ends at once — the iOS BLE stack and the scale's USB-C debug port (115200; use pyserial with `dtr=False, rts=False` so opening the port doesn't DTR-reset the ESP32).

An A/B of the **same firmware**, macOS vs iOS 9, was decisive:

| scale-side | macOS (works) | iOS 9 (fails) |
|---|---|---|
| `Device connected` (onConnect) | prints | **never prints** |
| `onSubscribe` (FFF4) | yes | yes |
| every write (LED-on, heartbeat) ACKed | yes | yes |
| `onStatus` on each `notify()` | `s=1` SUCCESS_NOTIFY | `s=5` **ERROR_NO_CLIENT** |
| client receives weight | 237 packets | **0** |

So on iOS 9 the GATT layer is fully up — the client connects, subscribes to FFF4, and every command is received and ACKed — but the **arduino-esp32 BLE server never runs its connect path**: `onConnect` does not fire and `getConnectedCount()` stays `0`. `BLECharacteristic::notify()` therefore bails at its `getConnectedCount()==0` check with `ERROR_NO_CLIENT`, and the scale measures weight but never transmits it.

(macOS's identical `setNotifyValue` works, which is why the scale streamed there. iOS 9's older CoreBluetooth is the only client that trips this.)

## The fix

`onSubscribe` *does* fire on iOS 9, so use it:

1. **Adopt the subscription as the live connection** when none was registered — `setBleFff4Connection(desc->conn_handle, desc->conn_handle)` from the subscribe callback (mirrors what `onConnect` does).
2. **Count a valid conn handle** in `bleHasLiveClient()` so `bleCanNotifyCurrent()` passes for that client.
3. **Send through `bleNotifyReadPacket()`** — the normal `notify()` for server-counted clients, and a direct `ble_gatts_notify_custom()` on the conn handle when the server has no counted peer (iOS 9). All read-notification senders route through it.

No behavior change for registered (macOS / modern iOS / Android) clients — they take the normal `notify()` path.

## Verification

Flashed to an HDS dev board (ESP32-S3) over USB and tested with de1app on a jailbroken iPad mini 1 (iOS 9.3.5):

- **iOS 9:** ~290 weight packets received; de1app watchdog no longer logs "abandoning" — confirmed live in de1app (scale weight updates in real use).
- **macOS (normal client):** unchanged (~180 packets), still via the standard `notify()` path.
- Builds clean on `env:esp32s3`.

### DIFF
--- a/docs/AI_PROTOCOL_NOTES.md
+++ b/docs/AI_PROTOCOL_NOTES.md
@@ -29,7 +29,7 @@ A live BLE connection does not imply an FFF4 notification subscription. A client
 
 All outbound FFF4 paths pass through `bleCanNotifyCurrent()`. The shared gate checks the feature state, characteristic, live connection, and current subscription before touching the characteristic.
 
-On iOS 9 the arduino-esp32 BLE server never runs its connect path: `onConnect` does not fire and `getConnectedCount()` stays zero, even though the client subscribes to FFF4 and its writes are acknowledged. `onSubscribe` adopts that subscription as the live connection, `bleHasLiveClient()` counts a valid connection handle, and every gated sender emits through `bleNotifyReadPacket()`: the server's `notify()` for counted clients, or a direct `ble_gatts_notify_custom()` on the connection handle when the server counts no peer.
+On iOS 9 the pinned arduino-esp32 framework can insert a peer before descriptor lookup, then skip both `onConnect` and its connection-count increment when that lookup fails. `onSubscribe` removes the uncounted peer before disconnect can underflow the framework counter, adopts the connection without clearing queued responses, and records that the connection requires direct NimBLE notifications. Normal connections continue through `BLECharacteristic::notify()`.
 
 Display status responses are deferred through a mailbox. The BLE callback queues the response, and `processBleStatusResponse()` drains it from the main loop. The mailbox waits two seconds for the FFF4 subscription, then rechecks the connection handle, connection generation, pending requests, and subscription before disconnecting. A newer request or reused connection cancels stale recovery.
 
--- a/docs/AI_PROTOCOL_NOTES.md
+++ b/docs/AI_PROTOCOL_NOTES.md
@@ -29,6 +29,8 @@ A live BLE connection does not imply an FFF4 notification subscription. A client
 
 All outbound FFF4 paths pass through `bleCanNotifyCurrent()`. The shared gate checks the feature state, characteristic, live connection, and current subscription before touching the characteristic.
 
+On iOS 9 the arduino-esp32 BLE server never runs its connect path: `onConnect` does not fire and `getConnectedCount()` stays zero, even though the client subscribes to FFF4 and its writes are acknowledged. `onSubscribe` adopts that subscription as the live connection, `bleHasLiveClient()` counts a valid connection handle, and every gated sender emits through `bleNotifyReadPacket()`: the server's `notify()` for counted clients, or a direct `ble_gatts_notify_custom()` on the connection handle when the server counts no peer.
+
 Display status responses are deferred through a mailbox. The BLE callback queues the response, and `processBleStatusResponse()` drains it from the main loop. The mailbox waits two seconds for the FFF4 subscription, then rechecks the connection handle, connection generation, pending requests, and subscription before disconnecting. A newer request or reused connection cancels stale recovery.
 
 ## Framing And Checksums
--- a/include/ble.h
+++ b/include/ble.h
@@ -305,6 +305,17 @@ class Fff4Callbacks : public BLECharacteristicCallbacks {
 #if defined(CONFIG_NIMBLE_ENABLED)
   void onSubscribe(BLECharacteristic *pCharacteristic, ble_gap_conn_desc *desc, uint16_t subValue) {
     if (desc == nullptr) return;
+    // iOS 9 CoreBluetooth establishes the link and subscribes but the arduino-esp32
+    // BLE server never runs its connect path (onConnect does not fire, getConnectedCount
+    // stays 0). onSubscribe still fires, so adopt this subscription as the live
+    // connection when none was registered.
+    if (subValue != 0 && connId == 0xFFFF) {
+      setBleFff4Connection(desc->conn_handle, desc->conn_handle);
+      t_firstConnect = millis();
+      t_heartBeat = millis();
+      bleState = CONNECTED;
+      deviceConnected = true;
+    }
     const uint16_t nextHandle = subValue == 0 ? 0xFFFF : desc->conn_handle;
     portENTER_CRITICAL(&bleFff4Mux);
     const bool changed = desc->conn_handle == connId && bleFff4SubscriptionHandle != nextHandle;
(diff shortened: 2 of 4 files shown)


ODevStudio (changes requested):
Thanks John for the PR 

I reviewed PR #146 against the current `de1app` Decent Scale flow and the pinned Arduino-ESP32 3.3.11 BLE implementation. The **first-session fix is sound**, but I would **request changes before merging** because there is a reconnect blocker in the underlying failure path.

### 1. Blocker: the exact iOS 9 failure path corrupts `getConnectedCount()` on disconnect

The project pins Arduino-ESP32 **3.3.11**.  In that version, the NimBLE `BLEServer` connect handler does this in order:

```cpp
server->m_connId = event->connect.conn_handle;
server->addPeerDevice(...);
rc = ble_gap_conn_find(event->connect.conn_handle, &desc);
if (rc != 0) {
    return 0;
}
// onConnect callbacks...
// m_connectedCount++
```

So the peer is inserted **before** `ble_gap_conn_find()`, but the counter is incremented only afterward.

That lines up extremely well with your observations. More importantly, Arduino's characteristic subscription handler also calls `ble_gap_conn_find()` and **will not call `onSubscribe()` if that lookup fails**.  Therefore:

1. iOS 9 GAP connect arrives.
2. `addPeerDevice()` succeeds.
3. `ble_gap_conn_find()` apparently races/fails at that instant.
4. `onConnect()` is skipped and count remains `0`.
5. A little later the CCCD subscription arrives.
6. `ble_gap_conn_find()` now succeeds, so `onSubscribe()` fires.

That strongly suggests the PR has identified the right symptom but not the full lifecycle consequence.

On disconnect, Arduino does:

```cpp
if (server->removePeerDevice(conn_handle, false)) {
    server->m_connectedCount--;
}
```

Because the peer was inserted but the count was never incremented, this is **0 → `UINT32_MAX`**.

That then breaks reconnect. The Arduino advertising implementation refuses to start advertising when:

```cpp
pServer->getConnectedCount() >= CONFIG_BT_NIMBLE_MAX_CONNECTIONS
```

Your `onDisconnect()` subsequently calls `pAdvertising->start()`, but after the underflow the library believes it has billions of connections.

So I expect this sequence to reproduce a failure:

```text
boot
→ iOS 9 connect
→ weights work with PR
→ disconnect iPad
→ scale does not advertise again
```

There is a second consequence. PR #146 changes:

```cpp
return pServer != nullptr &&
       (pServer->getConnectedCount() > 0 || connId != 0xFFFF);
```

After that underflow, `bleHasLiveClient()` remains true even after your own `connId` has been cleared. That is observable outside BLE: finger/button handling explicitly uses `bleHasLiveClient()` to decide whether local button behavior is allowed.

**I would treat this as blocking.**

The cleanest fix is upstream in Arduino-ESP32: don't insert the peer without balancing `m_connectedCount`, or move the count update before the descriptor lookup. But for an OpenScale-local workaround, you can undo the phantom peer entry when you detect this fallback:

```cpp
if (subValue != 0 && connId == 0xFFFF) {
    const bool uncounted = pServer->getConnectedCount() == 0;

    if (uncounted) {
        // Arduino BLEServer inserted this peer before its failed
        // ble_gap_conn_find(), but never incremented m_connectedCount.
        // Remove it so disconnect won't decrement 0 -> UINT32_MAX.
        pServer->removePeerDevice(desc->conn_handle, false);
    }

    ...
}
```

Then track that this connection requires direct notifications rather than consulting `getConnectedCount()` again. I would also make `bleHasLiveClient()` derive from `connId` plus `ble_gap_conn_find()` rather than the poisoned library counter.

### 2. The adoption currently destroys responses DE1app queued before subscribing

`setBleFff4Connection()` calls `resetBleFff4StateLocked()`, which clears both:

```cpp
bleStatusResponsesPending = 0;
bleVoltageResponsesPending = 0;
```

That is correct during a normal `onConnect()`, because the connection is registered before application traffic arrives. It is not correct when `onSubscribe()` is being used as a late substitute for `onConnect()`.

This matters specifically with DE1app. On a Decent Scale connection it does approximately:

```text
0 ms    heartbeat
200 ms  LED/display-on command
300 ms  enable FFF4 notifications
400 ms  enable FFF4 notifications again
500 ms  LED/display-on again
```

The LED/display command queues a status response in OpenScale; that mailbox was explicitly designed to wait for the notification subscription.  But at 300 ms PR #146 calls `setBleFff4Connection()` from `onSubscribe()`, which erases the response queued by the 200 ms command.

DE1app happens to mask this because it sends the LED command again at 500 ms. A client that sends the command only once would lose its response.

I'd introduce a separate adoption operation that **preserves the mailboxes**, for example conceptually:

```cpp
adoptBleFff4ConnectionFromSubscription(handle);
```

which updates:

```text
connection generation
connId
subscription handle
notification failure state
```

without resetting pending status/voltage responses.

The existing contract test doesn't catch this; its simulated mailbox assumes subscription merely sets the subscription field and preserves pending responses.

### 3. Direct notification failures are now invisible

The normal Arduino `notify()` path invokes `onStatus()` on failures such as `ERROR_NO_CLIENT`.  The new direct path does:

```cpp
struct os_mbuf *om = ble_hs_mbuf_from_flat(data, len);
if (om != nullptr)
    ble_gatts_notify_custom(connId, pReadCharacteristic->getHandle(), om);
```

and discards the return code.

Given that this code exists specifically for an anomalous BLE stack path, I would capture and rate-limit-log the `ble_gatts_notify_custom()` return value. Also log allocation failure. Otherwise a future failure becomes harder to diagnose than the bug this PR fixes.

### DE1app protocol compatibility itself looks correct

There is no wire-format problem here. DE1app enables notifications on the Decent Scale read characteristic and routes incoming notifications through `parse_decent_scale_recv`; 7-byte `0xCE` packets are interpreted as weight and passed into `process_weight_update()`.

OpenScale's `buildWeightPacket()` still produces the same 7-byte Decent Scale packet, and `bleNotifyReadPacket()` transmits those exact bytes.  So the architectural idea—**use the successful FFF4 subscription as evidence of the real connection and issue a raw NimBLE notification on that handle**—is correct.

I would add these hardware tests before merging:

* iOS 9 connect → verify continuous weight.
* iOS 9 disconnect → verify scale immediately advertises again.
* iOS 9 reconnect **without reboot** → verify weight again.
* Then connect macOS/modern iOS without reboot.
* Verify local tare/timer buttons behave normally after the iOS 9 client disconnects.
* Send LED-on once *before* CCCD subscription, then subscribe, and verify its queued status response is delivered.
* Log `getConnectedCount()` during the first iOS 9 disconnect; I strongly expect you'll see `4294967295` with the current framework.

So: **the notification workaround is directionally right, but I would not merge #146 as-is until the phantom peer / counter-underflow path is handled.** The counter issue is the one likely to turn a successful one-shot test into a failed real-world reconnect.

PR #163, opened 2026-08-24
Restore legacy USB weight-enable framing

ODevStudio:
## Summary
- prefer the current four-byte USB weight-enable command
- accept the legacy three-byte enable form after the existing 30 ms timeout
- retain immediate three-byte handling for disable commands

## Verification
- python tools/test_decent_weight_frame_contract.py
- pio run -e esp32s3


ODevStudio:
Not needed, so closing.

PR #165, opened 2026-08-27
Start a targeted firmware update from any transport

skialpine:
## Summary

- extend Decent opcode `0x1B` with an optional three-byte version payload so a connected app can name the release to install
- add a `wifi_update` control command to the `/snapshot` WebSocket, accepted bare, as `wifi_update <version>`, and as JSON
- install a named release without the on-OLED picker or the hold-to-confirm gesture; a request with no version keeps the interactive flow on every transport
- resolve a requested version against the picker's own selection list so the unattended path inherits every existing eligibility rule
- acknowledge a start request as queued, not as installable

## Context

An app connected over BLE or USB can already read the public release catalog and decide which release a scale is eligible for. What it cannot do is finish: `0x1B` carries no payload, so it drops the scale into the on-OLED picker and a user who already chose a release in the app is asked to choose it again on the hardware. Remote start is also unavailable over WiFi, even though `WSP_WIFI_UPDATE` already exists and the WebSocket already carries tare, timer, display, sleep, and power commands.

## Wire compatibility

Each version byte is encoded `0x80 | value`, capped at 127. This is load-bearing rather than cosmetic. `processUsbRxBuffer()` enters the binary framer only when the buffer's first byte is `0x03`, so a payload byte can be mistaken for a command only if it equals `0x03`. Setting the high bit puts every version byte in `0x80..0xFF`, which the framer can never read as a frame start and no ASCII text command can match.

That yields one request form that is correct against every scale, with no capability negotiation and no version gate in the client:

- firmware predating this change frames `03 1B` as two bytes and starts its picker, the correct graceful fallback, then discards the three trailing bytes through the text path
- this firmware disambiguates on `data[2] >= 0x80`; a following command always begins `0x03`, and a biased version byte never can, so the forms are distinguished with no checksum and no reliance on the 30 ms frame timeout
- an existing client sending a lone `03 1B` still returns 2 immediately, with no added latency

An unbiased payload would not be safe: with `major = 3`, a `3.10.2` target decodes to `03 0A 02`, the power-off command, on older firmware, and `3.11.x` lands on the timer opcode. This is recorded in the compatibility checklist in `docs/AI_PROTOCOL_NOTES.md`.

## Safety

The unattended path skips only `pullOtaPickRelease()` and `pullOtaConfirmInstall()`. Signature verification, rollback-manifest resolution, HTTPS CA validation, asset URL allowlisting, exact size checks, SHA-256 verification, and the staged LittleFS transaction are shared and unchanged.

`pullOtaRunUpdate()` resolves a target against `selection` rather than the raw catalog, so it inherits every rule `pullOtaAddParsedRelease()` and `pullOtaBuildSelectableReleases()` already applied. A release the picker would not have offered is simply not found and the request is refused, without falling back to the picker. The client contributes a version number and nothing else; every asset URL, size, and hash comes from the scale's own signature-verified fetch.

Accept-time refusals cover a malformed version and an already queued or running update. Catalog-level refusals reach `pullOtaFail()` on the OTA task, after `stopWebServer()` has closed WebSocket clients, so they surface on the display and serial log only. No client progress stream is added.

## Validation

```
pio run -e esp32s3
pio run -e esp32s3 with -DHDS_FEATURE_PULL_OTA=0
pio run -e esp32s3-grinder
pio run -e esp32s3-energy-menu
python tools/test_pull_ota_contract.py
python tools/test_decent_ota_target_contract.py
python tools/test_ota_rollback_contract.py
python tools/test_ota_reboot_routing_contract.py
python tools/test_ota_input_isolation_contract.py
python tools/test_ota_public_key_header.py
python tools/test_generate_release_manifest.py
python tools/test_release_workflow_contract.py
python tools/test_ble_subscription_contract.py
python tools/test_usb_text_actions_contract.py
python tools/test_usb_text_buffer_contract.py
python tools/test_soft_sleep_ads_wake.py
python tools/test_ai_docs_contract.py
```

No new build warnings. Each commit builds and passes the suite on its own.

## Hardware verification

Verified on a scale over USB with WiFi connected.

Against firmware predating this change, a versioned frame degraded exactly as designed, with no power-off, timer, or other side effect:

```
Received HEX: 031B
Start WiFi OTA
Received HEX: 83818D
handling <unprintable>
```

Against this change:

| Case | Result |
| --- | --- |
| `03 1B` followed by `03 22` in one write | `Received HEX: 031B` then `Received HEX: 0322` |
| `03 1B 83 81 8D`, already-installed 3.1.13 | `[pull-ota] error Version not offered`, no picker |
| `03 1B 83 81 E3`, absent 3.1.99 | `[pull-ota] error Version not offered`, no picker |
| `03 1B 83`, incomplete payload | `Dropping incomplete USB binary frame`, no bare dispatch |
| WebSocket malformed versions, text and JSON | `ota_version_invalid` |
| WebSocket two requests before dispatch | `status ok` then `ota_busy` |
| WebSocket `wifi_update 3.1.99` | `status ok`, queued, then `Version not offered` |
| WebSocket bare `wifi_update` | `status ok`, queued, interactive path |

Not verified: a completed unattended install. The hardware run above was made against a `3.1.13` tree, where the only catalog release was the version the scale already ran, so nothing was installable. Everything up to and including release resolution was exercised; the write path below `pullOtaInstall()` is unchanged shared code.

After the rebase onto `3.1.14-preview.1` this is testable, because stable `3.1.13` is now an eligible signed downgrade rather than the installed version. That run has not been made yet.

## Rebase notes

Rebased onto `main` after the framing helpers were extracted into `include/decent_protocol_frame.h`. The `0x1B` length rule moved there with them, and `decent_protocol.h` keeps only the dispatch change. `pull_ota_target.h` depends on nothing beyond `stddef`, `stdint`, and `stdio`, so the frame header still compiles for the native suite.

The framing rule is now covered by native cases in `test/test_decent_protocol` rather than only by a source contract: a bare request frames as two bytes, a biased payload as five, an incomplete payload returns 0, a following command is not swallowed, and biased bytes decode to the right version. The source contract additionally asserts the framer is not copied back into `decent_protocol.h`.

The explicit `nightly.yml` test entry was dropped; that workflow now loops over `tools/test_*.py` and discovers the new contract on its own.

## Follow-up, not in this change

`buildLedResponsePacket()` packs the reported version as `verLow = (minor << 4) | patch`, one nibble each, so `3.1.16` would report as `3.2.0` over BLE and USB. This change does not introduce it and the new payload uses a full byte per component, but the installed version that both an app's eligibility check and the already-installed refusal depend on is read through that packet. It should be fixed before a `3.1.16` release, as its own change.


ODevStudio (changes requested):
Thanks for the PR. The overall approach looks good: resolving the requested version through the existing selectable-release list keeps OTA eligibility rules centralized, and the install path continues to reuse the existing verification and rollback machinery.

I found two correctness issues that should be addressed before merge.

1. USB framing can consume the next command
In decentCommandFrameLength(), once data[2] is biased, the parser waits for a five-byte 0x1B frame. However, the remaining payload bytes are not validated before those five bytes are consumed.

For example:

03 1B 83 followed by 03 22

can become:

03 1B 83 03 22

The parser then treats the complete sequence as the OTA request, decodes 03 22 as version bytes, and loses the following command.

Please make the framing logic reject an incomplete targeted payload when a new 0x03 frame boundary appears rather than consuming it as version data.

It would be good to add regression tests for:

03 1B 83 + 03 22
03 1B 83 81 + 03 22
2. OTA request acceptance has a dequeue/dispatch race
processWsPendingCmds() clears WSP_WIFI_UPDATE when it snapshots the target, but b_ota / b_pullOtaRunning are only set later when wifiUpdate() is called.

That creates a short window where another WebSocket request can observe:

no pending WSP_WIFI_UPDATE
no OTA currently marked as running
and therefore receive status ok.

The second accepted request may then either run after the first failure or disappear if the first OTA succeeds and reboots.

Please keep the OTA request in a queued/dispatching state until the first request has transitioned to running, so acceptance is atomic across that handoff.

skialpine:
Thanks — both fixed in 910deb9 and 9a8419c.

**1. USB framing**

Every byte of a started payload must now carry the `0x80` bias. On an
unbiased byte the resolver consumes one byte and resynchronizes instead of
framing five, so the truncated request is refused and the next command
survives: `03 1B 83` + `03 22` and `03 1B 83 81` + `03 22` both dispatch
`03 22` with no update started. `03 1B 83 81` alone still waits. I refused
rather than falling back to the bare form, which would have started the
picker for a request the client never made.

The dispatcher repeats the check before decoding, since BLE and the
WebSocket deliver whole packets and never reach the USB resolver.

Regression tests for both cases plus the waiting case are in
`test/test_decent_protocol/test_main.cpp` and
`tools/test_pull_ota_target_contract.cpp`.

**2. Acceptance race**

Added `volatile bool pendingOtaDispatching` under `wsPendingMux`.
`processWsPendingCmds()` sets it in the same critical section that clears
`WSP_WIFI_UPDATE`, and `remoteQueueWifiUpdate()` refuses while it is set, so
a request arriving mid-handoff gets `ota_busy`. It is cleared only after
`wifiUpdate()` returns with the running flags set, or alongside the pending
bit when the request is deferred. Lock order is unchanged.

Docs updated in `docs/AI_PROTOCOL_NOTES.md` and `docs/AI_OTA_NOTES.md`.
Verified: native tests 12/12, host contract and all OTA contract scripts
pass, `pio run -e esp32s3` builds clean.

ODevStudio (changes requested):
Thanks for the updates. The two issues from my previous review are resolved.


USB framing swallowing a following command and the queued-to-dispatch acceptance race — look addressed in the current head.

Another review round still found 2 issues.

1. **USB can overtake an already queued BLE/WebSocket OTA request**

`remoteQueueWifiUpdate()` now correctly prevents a second queued request while `WSP_WIFI_UPDATE` is pending or `pendingOtaDispatching` is set.

USB, however, still bypasses that admission path and calls `::wifiUpdate(target)` directly after checking only `b_pullOtaRunning || b_ota`.

That leaves a case like:

* `processWsPendingCmds()` runs and finds no OTA request
* a BLE/WebSocket request arrives and queues `WSP_WIFI_UPDATE`
* USB receives `03 1B ...` later in the same loop iteration
* USB sees no OTA running yet and starts its update directly

The previously accepted BLE/WebSocket request remains queued. If the USB update later fails or exits without rebooting, that queued request can run afterward; if the USB update succeeds and reboots, the accepted queued request disappears.

This breaks the documented invariant that only one start request exists at a time.

It would be better for USB and the queued transports to share the same atomic OTA admission/reservation mechanism, so a pending request blocks a direct USB start as well.

A regression test covering “queued remote request followed by direct USB start before dispatch” would be useful.

2. **A non-string JSON `action` is treated as a bare `wifi_update`**

`handleWebsocketNamedJsonCommand()` currently maps both a missing `action` and an `action` of the wrong JSON type to `""`.

For most commands that may be harmless, but for `wifi_update` an empty action has meaning: it starts the bare interactive OTA flow.

For example:

`{"command":"wifi_update","action":false}`

currently behaves like:

`{"command":"wifi_update"}`

rather than being rejected as malformed.

Since a present-but-invalid `action` is different from an omitted action, I think this should return `ota_version_invalid` or another malformed-command error instead of starting OTA.

A test for boolean, numeric, and object `action` values would cover this boundary well.

Apart from those two points, the targeted-release resolution and the fixes from the previous review look good.

skialpine:
Both fixed in d8b8215 and 2d2953c, verified on a scale.

1. USB now admits through `remoteQueueWifiUpdate()` like BLE and the
   WebSocket, so a queued request can't be overtaken. A request that reaches
   dispatch while an update is running is dropped rather than requeued.

2. A present JSON `action` that isn't a string answers `invalid_action`;
   omitted or `null` still means the bare form. Boolean, numeric, and object
   cases added to `tools/ws_command_test.py` (32/32 against a live scale).

PR #190, opened 2026-09-08
Improve custom build installation status reporting

ODevStudio:
## Implemented
- Poll Fleet overview every 30 seconds while an assignment remains unconfirmed.
- Pause requests for hidden tabs, active edits and in-flight requests; stop when installed and desired hashes match.
- Prevent stale background responses from overwriting newer changes and explicitly bypass the browser request cache.

## Still required before review
- Add download-progress reporting, clearly separate from confirmed installation. ZIP/USB downloads must not imply that a particular scale installed the build.
- Add an automatic installed-build check-in after successful firmware and LittleFS boot, including USB installs when the paired scale has WiFi connectivity.
- Keep offline installations unconfirmed until actual device confirmation arrives.
- Add focused tests and complete firmware validation for those additions.

## Validation so far
- 17 existing configurator tests and 2 polling tests passed locally.
- Diff whitespace check passed.
- New polling has not been deployed; firmware/download reporting is not implemented yet.

Do not merge automatically. User requested review before merging.

### DIFF
--- a/.github/workflows/custom-build-configurator.yml
+++ b/.github/workflows/custom-build-configurator.yml
@@ -35,3 +35,4 @@ jobs:
           node --check docs/custom-build/app.js
           node --check docs/custom-build/fleet.js
           node --test cloudflare/custom-build-worker/test/configurator.test.mjs
+          node --test cloudflare/custom-build-worker/test/fleet-polling.test.mjs
--- a/.github/workflows/custom-build.yml
+++ b/.github/workflows/custom-build.yml
@@ -223,8 +223,8 @@ jobs:
           ref: ${{ inputs.builder_commit || github.sha }}
           fetch-depth: 0
           persist-credentials: false
-      - name: Create prospective release tag for main validation
-        if: inputs.firmware_ref == 'main'
+      - name: Create prospective release tag for development validation
+        if: inputs.firmware_ref == 'main' || inputs.firmware_ref == 'v3.1.14-preview.3'
         run: |
           if ! git show-ref --verify --quiet refs/tags/v3.1.14; then
             git tag v3.1.14 "${{ github.sha }}"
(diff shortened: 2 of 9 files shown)


tadelv (changes requested):
Findings
P1 — CI is broken by the fleet.js cache-version bump.
app.js now imports fleet.js?v=12, but tools/test_plugin_catalog.py still asserts fleet.js?v=11.
This is already failing the Custom build configurator → Validate configurator contracts job.
Fix: update that contract assertion to fleet.js?v=12.
P2 — app.js itself needs a cache-key bump.
The PR changes app.js from importing fleet v11 to v12, but index.html still serves it as app.js?v=27, exactly the same URL as main.
That defeats the cache-busting chain for clients holding the existing app.js?v=27: they can continue running the old app module and consequently load fleet.js?v=11. Previous configurator changes explicitly advanced this outer version—for example, the preceding version was app.js?v=26.
Fix: bump the entry point to app.js?v=28 and update the corresponding contract test.
P2 — the new polling tests are not run in CI.
The PR adds fleet-polling.test.mjs with two tests, but the dedicated configurator workflow runs only:
node --test .../configurator.test.mjs.
The broader custom-build workflow runs configurator.test.mjs, worker.test.mjs, and fleet.test.mjs, but likewise omits the new polling suite.
Fix: add node --test cloudflare/custom-build-worker/test/fleet-polling.test.mjs to at least the configurator workflow, preferably both relevant contract jobs.
Implementation assessment

The actual race fencing looks sensible: mutations increment generation, overlapping overview responses are discarded, activeRequests prevents polling during in-flight work, and background results are discarded when the page becomes hidden or editing starts. I don't see a blocker in that mechanism itself.

However, this is still explicitly a draft. The PR description says automatic post-install device check-in, separate download progress, offline confirmation semantics, and focused firmware validation are still outstanding. So even after the three issues above are fixed, I would keep it unapproved until those stated requirements land.

ODevStudio:
Fix from this review/comment
Verify the three review fixes in commit 76af446: the fleet.js?v=12 contract, the app.js?v=28 entry point and matching assertion, and fleet-polling.test.mjs in the configurator workflow and both custom-build contract jobs. Run the cache-transport test and relevant Node tests. Correct any remaining mismatch in those three fixes; if they already pass, report that no further change is needed. Keep the broader draft firmware work outside this check. Do not commit, push, or post another comment.

Issue #196, opened 2026-09-10
[Feature Request] USB "enob" command for ADC effective-resolution testing


Sofronio:
Reading the ADC noise floor on real hardware currently requires either flashing a temporary test firmware or attaching a logic analyser. I want a built-in way to measure the front-end quality on the bench, without rebuilding anything.

I want to add a USB serial command that samples raw 24-bit ADS1232 values and prints a statistical report:

- **`enob`** — sample 100 values and print the statistics
- **`enob <count>`** — sample a custom count (10–1000)

The report contains: effective SPS, average, standard deviation, peak-to-peak drift, ENOB (Effective Number of Bits) in both p-p and RMS terms, and the equivalent resolution in mg derived from the current calibration factor — with a note that the mg figures are only meaningful with a correct calibration. The final assessment line is graded by the RMS resolution in mg (the figure that matters for weighing), so the verdict is comparable across units with different calibration factors.

Sampling is hardware-paced, so the command works unchanged on both 10 SPS and 80 SPS hardware. A 1 s settle precedes collection so a keypress or touch does not disturb the measurement.

Use case: verifying the noise floor after a hardware change (reference, layout, LDO), comparing boards, and checking whether a firmware change affected the analog front-end.

PR #197, opened 2026-09-10
feat(usb): add "enob" command for ADC effective-resolution testing

Sofronio:
Adds a USB serial command that measures the analog front-end quality on real hardware, without flashing test firmware or attaching an analyser.

Implements #196.

## Usage

```
enob                sample 100 raw values
enob 500            sample a custom count (10-1000)
```

Sampling is hardware-paced, so the command works unchanged on both 10 SPS and 80 SPS hardware. A 1 s settle period precedes collection so a keypress or touch does not disturb the measurement.

## Report

- effective SPS (measured, not assumed)
- average and standard deviation of the raw 24-bit values
- peak-to-peak drift
- ENOB (Effective Number of Bits) in both p-p and RMS terms
- equivalent resolution in mg, derived from the current calibration factor, with a note that these figures are only meaningful when the scale is correctly calibrated
- min/max raw, total sample count, and an assessment line graded by the RMS resolution in mg (≤50 mg reads GOOD), so the verdict is comparable across units with different calibration factors

## Implementation notes

- New `include/enob_test.h`: self-contained, samples through the existing polling API (`scale.update()`), so it does not conflict with the AsyncTCP task or the normal sampling model.
- The command lives in the existing USB command dispatch (`include/usbcomm.h`) next to the other bench commands.
- No change to normal firmware behaviour; the test only runs when the command is issued.

## Testing

- `esp32s3` build green; docs/contract checks green.
- Run on hardware at 10 SPS: ~16.4 ENOB (p-p), ~100 mg p-p resolution / ~19 mg RMS; repeated runs are consistent.
- Verified the mg-based assessment on a noisier unit (35 mg RMS → GOOD), confirming the grading tracks the resolution figure rather than raw LSB counts.

### DIFF
--- a/include/enob_test.h
+++ b/include/enob_test.h
@@ -1,0 +1,204 @@
+#ifndef ENOB_TEST_H
+#define ENOB_TEST_H
+
+#include <Arduino.h>
+
+// ENOB (effective resolution) test — samples raw 24-bit ADC values and
+// prints noise, resolution and noise-free-bit statistics.
+// Uses polling mode (scale.update()) so it does NOT conflict with the
+// AsyncTCP task or the existing sampling model.
+
+static void runEnobTest(const char *mode, int sampleCount = 100) {
+  Serial.println();
+  Serial.println("=============================================");
+  Serial.print("=== ENOB (Effective Number of Bits) Test: ");
+  Serial.print(mode);
+  Serial.println(" ===");
+  Serial.println("Hardware-paced, SPS measured below");
+  Serial.println("=============================================");
+
+  // Allocate on heap — ESP32 has plenty of PSRAM
+  long *rawData = (long *)malloc(sampleCount * sizeof(long));
+  if (rawData == NULL) {
+    Serial.println("ERROR: Failed to allocate sample buffer");
+    return;
+  }
+
+  unsigned long t_start = millis();
+  int collected = 0;
+
+  while (collected < sampleCount) {
+    // scale.update() returns 1 when a fresh conversion was read; the ADC
+    // paces the sampling, so 10 SPS and 80 SPS hardware both work as-is.
+    if (scale.update()) {
+      rawData[collected] = scale.getDebugInfo().rawValue;
+      collected++;
+    }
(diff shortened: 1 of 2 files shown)


ODevStudio (changes requested):
Thanks for adding this. The command fits the existing polling path and gives us a useful bench test without a separate firmware build.

I found a few issues I think we should address before merge.

**ENOB calculation**

`noiseFreeBits` uses `0x800000` (`2^23`) as the full-scale code span. For a 24-bit bipolar code range, that puts the p-p result about one bit low.

The RMS calculation also multiplies `stdDev` by `6.6`, which converts RMS noise to a p-p estimate. If we want to report both metrics, I suggest:

* noise-free resolution (p-p): `24 - log2(p2pCodes)`
* effective resolution (RMS): `24 - log2(rmsCodes)`

The names would also make the distinction clearer in the output.

**ADC timeout**

The collection loop can run forever if the ADC stops producing conversions. Broken wiring or a stalled ADC is a case this diagnostic should handle.

Could we abort on the existing signal-timeout flag, or add a test timeout and free the buffer before returning?

**Saturated readings**

The driver exposes `dataOutOfRange`, but the test only records `rawValue`. An ADC stuck at full scale can produce `stdDev == 0` and receive a `GOOD` assessment.

I’d mark the measurement invalid if we see an out-of-range sample.

**Uncalibrated assessment**

The report detects `1000.0f` as the default, uncalibrated factor, then uses that same value to calculate the mg resolution and assessment.

I think the command should print `UNKNOWN` and skip the mg-based grade until the scale has a real calibration factor.

**SPS calculation**

`sampleCount / elapsed` includes the wait for the first conversion, which biases short runs.

Recording the first and last sample timestamps and using `(N - 1) / elapsed` would measure the conversion rate without that initial wait.

**Small cleanup**

`stdDev * 1000` is milli-LSBs, not micro-LSBs.

For the settle timer, `millis() - start < 1000` would also handle `millis()` rollover safely.

The command will be useful for checking front-end changes on real hardware. I’d fix the measurement formulas and invalid-hardware cases before merging, then add host-side tests for the formulas, timeout, saturation, and default-calibration case.

PR #203, opened 2026-09-10
Allow saving builds before pairing a scale

ODevStudio:
## Summary

- Allow the wizard's Save build action before pairing a scale. Generate and persist a recovery key on first save, then use the normal fleet library API.
- Create the library transactionally only for a verified ready build. Preserve saved builds when the first scale is paired; never assign or install automatically.
- Reuse the existing bounded pairing-creation quota for new libraries and forward the edge-derived client key. Keep existing-library saves independent of the creation quota.
- Do not activate a new recovery key if local persistence fails; update browser cache versions and the operations notes.

## Validation

- All 46 Node configurator, fleet, Worker, polling, and retention tests passed.
- New browser-handler coverage: no scales, new/existing key, blocked storage, failed request, and build not ready.
- New API coverage: save before pairing, duplicate save, fleet isolation, preservation after pairing, no automatic assignment, missing-build rejection, and creation rate limiting.
- Plugin catalog and AI documentation contracts passed; JavaScript syntax and git diff --check passed.

Deploy the Worker before the new browser assets. This change does not modify firmware or release assets.

### DIFF
--- a/cloudflare/custom-build-worker/src/fleet.mjs
+++ b/cloudflare/custom-build-worker/src/fleet.mjs
@@ -411,10 +411,10 @@ async function addBuild(request, storage, env) {
   const ownerFleetId = await fleetId(request);
   const body = requireObject(await readJson(request), ["combination_hash"]);
   const key = `fleet:${ownerFleetId}`;
-  if (!(await storage.get(key))) throw new FleetError(404, "fleet_not_found");
   const reference = await readyBuildReference(env, body.combination_hash);
   return storage.transaction(async transaction => {
     const fleet = await transaction.get(key);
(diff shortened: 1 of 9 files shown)


ODevStudio:
Deployed after all 8 applicable CI checks passed. Worker version: 0ffba061-3744-4889-b8ba-0a9d6ae15642. Pages run 34511278993 succeeded; public HTML loads app.js?v=31 and fleet.js?v=14 with the pairing gate removed. Playwright checks at 1440px and 390px verified wizard save with zero paired scales, durable recovery key, reload persistence, and disabled assignment using mocked API responses. A non-writing production probe confirmed a new library request reaches build-readiness validation rather than fleet_not_found. No scales were paired or assigned and no production builds were created by verification.

PR #209, opened 2026-09-21
draft: add Shot Flow espresso extraction web app

TooLitToGit:
## Feedback Requested

I am opening this as a draft to understand whether Shot Flow fits the project and how standalone web applications should be contributed. _(Note that it could be renamed considering other non espresso use cases such as pour over, could call it extraction flow for example)_

Before refining it further (certainly it's nowhere near complete right now), I would appreciate guidance on:

1. Is this type of web application suitable for inclusion in this project?
2. What is the preferred contribution model and location for web applications?
3. Is there a separate submission or review process for isolated web applications that do not require firmware changes?

I would also love to contribute to broader improvements across the built-in web applications, such as:

- Unifying the existing CSS themes and shared visual components
- Adding consistent dark-mode support across all built-in web applications
- Improving the GitHub Pages landing page and navigation

I would appreciate guidance on whether these improvements are wanted and whether they should be proposed as separate focused pull requests.

I am happy to adjust Shot Flow to follow the maintainers’ preferred direction. Until that direction is clear, I will keep this PR in draft and continue development on my fork.

## Summary

This draft adds Shot Flow, a scale-hosted web application for visualizing espresso extraction data in real time.

**Demo:** https://toolittogit.github.io/openscale/shot-flow/shot_flow.html

## Potential Uses

Shot Flow could help users:

- Dial in an espresso machine by comparing beverage flow, yield, and extraction time between shots
- Identify whether a shot runs ahead of or behind a chosen target profile
- Evaluate grind changes using repeatable flow and yield measurements
- Save successful extractions as reusable target profiles
- Share flow curves and recipes with other users
- Match another user’s profile as a repeatable starting point for reproducing a preferred taste
- Compare different coffees, grinders, baskets, doses, and machine settings
- Build a shared library of profiles associated with particular coffees or desired flavor outcomes
- Move profiles between OpenScale and DE1 tooling through a documented interchange format

Flow matching cannot guarantee identical taste because water chemistry, coffee, roast age, grinder, basket, temperature, pressure, and puck preparation also affect extraction. The profiles are intended as measurable and shareable reference points rather than universal recipes.

## Features

- Live beverage flow, cumulative yield, and extraction-time graph
- Automatic shot start and stop detection with manual controls
- Dynamic bounds for longer or higher-yield extractions
- Dose-scaled Espresso and Turbo target profiles
- Free-pour mode without a target profile
- Target curves derived from published Decent profiles
- Optional grind-adjustment guidance
- Saving of actual measured shot samples
- Worker-backed IndexedDB persistence
- Management page for shots, profiles, and user settings
- Reusable target profiles created from saved shots
- Canonical `openscale.shot-profile/v1` format
- Bidirectional OpenScale and DE1 legacy Tcl conversion
- Hardware-free GitHub Pages demo using isolated mock data

## Implementation

The application is contained under:

`plugins/default-web-apps/assets/shot_flow/`

It uses the existing `/snapshot` WebSocket endpoint at 10 Hz and sends the existing `tare` command. No firmware behavior or hardware callback paths were changed.

Demo data is kept outside production web assets and added only while staging the GitHub Pages artifact.

The default web-app package remains within its existing LittleFS budget.

## Validation

- Browser-tested on desktop and mobile viewports
- Verified automatic shot lifecycle and graph rendering
- Verified target flow/yield integration
- Verified dynamic graph bounds
- Verified saved shots contain actual measurements rather than target data
- Verified IndexedDB persistence and deletion
- Verified DE1 conversion using published profiles
- Verified demo fixtures do not enter scale filesystem assets
- Plugin catalog, launcher, gzip, and whitespace contracts pass


ODevStudio:
Thanks for opening this as a draft. This kind of standalone web application is a good fit for OpenScale.

I also agree that having a more consistent visual style across the web applications would be useful. Some of the existing pages came from older development work and were created at different times, so the styling is currently inconsistent.

I will likely align the default web pages myself over the next couple of days around the style currently used by the Custom Build configurator:

https://decentespresso.github.io/openscale/custom-build/

Once that work is done, you are welcome to follow the same style for Shot Flow so the applications feel like one suite. It should not necessarily be a strict requirement for contributed applications, though — a plugin can still use its own visual style where that makes sense.

For the contribution model, standalone applications should generally be represented as normal Custom Build plugins under:

```text id="rry4xc"
plugins/<plugin-id>/
```

For Shot Flow that could look roughly like:

```text id="5l39hq"
plugins/shot-flow/
├── plugin.json
└── assets/
    └── shot_flow/
        ├── shot_flow.html
        ├── main.js
        ├── styles.css
        ├── manage.html
        ├── manage.js
        └── modules/
            └── ...
```

The purpose of the Custom Build system is to let a user create a **customized scale build with selected features, plugins and web applications**.

The plugin's HTML/JS/CSS is therefore installed into the scale's LittleFS filesystem and served by the scale itself.

This is separate from the OpenScale GitHub Pages site.

In particular, a plugin contribution should not add a new public application under:

```text id="9tobmn"
decentespresso.github.io/openscale/<plugin>/
```

and should not modify the upstream GitHub Pages deployment workflow as part of installing a plugin.

So the additional GitHub Pages workflow, staging scripts and upstream Shot Flow demo are not needed for the plugin contribution.

You are of course free to keep a GitHub Pages demo on your own fork while developing the application or to make it easier for reviewers to see what you are working on. That is useful during development, but it should remain separate from the production plugin that gets merged upstream.

For discovery inside Custom Build, I think it may make sense in the future for me to provide a lightweight preview mechanism instead of hosting complete runnable demos on the OpenScale website.

For example, we could eventually extend the plugin/catalog metadata so a plugin can provide things such as:

- a JPG/WebP screenshot or preview image;
- a link to its documentation;
- a short README or feature description;

That would let a user see what an application looks like and understand what it does before including it in a custom scale build, without turning the OpenScale GitHub Pages site into a hosting platform for every plugin.

The current plugin manifest does not yet have dedicated preview-image or documentation-link fields, so that would be a separate improvement to the Custom Build/catalog UI rather than something Shot Flow needs to implement itself in this PR.

For Shot Flow itself, since it uses the existing `/snapshot` WebSocket and existing scale commands, it should be an asset-only plugin and should not need a firmware patch.

A minimal manifest would be along these lines:

```json id="620e0u"
{
  "schema": 2,
  "id": "shot-flow",
  "name": "Shot Flow",
  "description": "Visualize extraction flow, yield, and time from the scale.",
  "tooltip": "Adds the Shot Flow web application to the scale.",
  "version": "1.0.0",
  "firmware_refs": ["main"],
  "requires": ["littlefs", "websocket"],
  "depends_on": [],
  "conflicts": [],
  "recommends": {
    "features": [],
    "plugins": []
  },
  "patches": {},
  "assets": [],
  "budget": {
    "firmware_flash_bytes": 0,
    "static_ram_bytes": 0,
    "littlefs_bytes": 0
  }
}
```

The exact dependencies can be adjusted depending on how we want to handle common web resources.

It is completely valid for one plugin to depend on another. For example, Shot Flow could use:

```json id="p0a2zv"
"depends_on": ["default-web-apps"]
```

and then reuse resources such as:

```text id="jjyal8"
/shared/reconnecting-websocket.js
```

from `default-web-apps`.

That is supported by the current Custom Build system. The restriction is only that two selected plugins may not both install a file to the same LittleFS target path. Reusing a resource owned by a dependency is fine; installing another copy to the same target is not.

Longer term, if we end up with several independently selectable web applications, we may decide to move common CSS/JS resources into a smaller shared web-assets plugin. I would not consider that necessary for Shot Flow right now.

Alternatively, if we decide that Shot Flow should become part of the standard application bundle rather than an optional application, then its current location under:

```text id="w51l3n"
plugins/default-web-apps/assets/shot_flow/
```

is structurally valid. In that case it becomes part of `default-web-apps` rather than its own selectable plugin.

My preference for contributed applications would be to start them as separate plugins. That keeps the default bundle curated and lets users decide which applications they want in their custom scale build. Applications that later become broadly useful can always be considered for inclusion in `default-web-apps`.

I will handle the alignment of the existing default web pages and shared styling separately, so you do not need to include those changes in this PR.

For an asset-only web application contribution, the main expectations would therefore be:

- place the application under `plugins/<plugin-id>/`;
- provide a `plugin.json` containing the actual feature/plugin dependencies;
- list only firmware refs that have actually been tested;
- measure the LittleFS usage and provide a reasonable budget;
- avoid firmware patches unless the application genuinely requires new firmware behavior;
- reuse shared resources through plugin dependencies where appropriate;
- regenerate `docs/custom-build/catalog.json` and `docs/custom-build/service-catalog.json`;
- run the existing plugin/catalog/custom-build validation;
- document what has actually been tested on hardware versus only in the browser;
- keep mock/demo data outside the production plugin assets;
- do not modify the upstream GitHub Pages deployment as part of the plugin.

There is no separate submission mechanism needed for an isolated web application. A normal PR containing the plugin, regenerated catalogs, relevant documentation/tests and validation results is the intended contribution path.

`docs/plugin-development.md` describes the general plugin model, and `plugins/default-web-apps/` is the current reference for an asset-only web application package.

For this draft, the main decision is therefore whether Shot Flow should be:

1. dependent on `default-web-apps`, or
2. its own `plugins/shot-flow/` Custom Build plugin.

For new contributed applications I would prefer the second model, while still allowing them to depend on and reuse resources from `default-web-apps`.

